PR #33, opened 2024-10-04
messages and tap backs

AndrewCLu:


apps/frontend/src/lib/message/index.ts
```diff
@@ -0,0 +1,174 @@
...
+    receiverEncryptionPublicKey,
+    senderSignaturePublicKey
+  );
+
+  // Encrypt the signature
+  const senderEncryptedSignature = await encrypt(
+    senderEphemeralEncryptionPrivateKey,
+    receiverEncryptionPublicKey,
+    signature
+  );
```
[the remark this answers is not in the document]
AndrewCLu: We're using ECDH, so both private secret + public key are needed to generated the shared secret, which is then used to AES encrypt the actual message

apps/backend/src/lib/controller/postgres/prisma/message/message.ts
```diff
@@ -0,0 +1,45 @@
...
+  );
+};
+
+PrismaPostgresClient.prototype.CreateMessages = async function (
+  messages: CreateMessageData[]
+): Promise<void> {
+  await this.prismaClient.message.createMany({
+    data: messages.map((message) => ({
+      ...message,
+      createdAt: new Date(),
```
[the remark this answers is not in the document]
AndrewCLu: good catch, fixed

apps/backend/src/lib/controller/postgres/prisma/message/message.ts
```diff
@@ -0,0 +1,46 @@
+import { PrismaPostgresClient } from "@/lib/controller/postgres/prisma/client";
+import { MessageData, MessageDataSchema, CreateMessageData } from "@types";
+
+PrismaPostgresClient.prototype.GetMessagesForUser = async function (
+  userSignaturePublicKey: string,
+  lastMessageFetchedAt: Date | undefined
+): Promise<MessageData[]> {
```
stevenelleman, line 7: What happens if there are no messages? Does `findMany` default to empty array or is it undefined / null?

apps/backend/src/lib/controller/postgres/prisma/message/message.ts
```diff
@@ -0,0 +1,46 @@
...
+  return prismaMessages.map((prismaMessage) =>
+    MessageDataSchema.parse({
+      receiverSignaturePublicKey: prismaMessage.receiverSignaturePublicKey,
+      senderEphemeralEncryptionPublicKey:
+        prismaMessage.senderEphemeralEncryptionPublicKey,
+      encryptedData: prismaMessage.encryptedData,
+      senderEncryptedSignaturePublicKey:
+        prismaMessage.senderEncryptedSignaturePublicKey,
+      senderEncryptedSignature: prismaMessage.senderEncryptedSignature,
+      createdAt: prismaMessage.createdAt,
```
stevenelleman, line 29: Could clean this up with new schema: 
`export const MessagesSchema = z.array(MessageDataSchema);`

apps/backend/src/routes/message/index.ts
```diff
@@ -0,0 +1,79 @@
...
+      if (!user) {
+        return res.status(401).json({ error: "Invalid authentication token" });
+      }
+
+      await controller.CreateMessages(messages);
+
+      return res.status(201).json({});
+    } catch (error) {
+      console.error("Error sending message:", error);
+      return res.status(400).json({ error: errorToString(error) });
```
stevenelleman, line 74: 500? And is there any validation we can do on our end to ensure messages are all valid? That would help distinguish client-side vs server-side errors.

apps/frontend/src/pages/people/[username].tsx
```diff
@@ -198,6 +213,25 @@ const UserProfilePage: React.FC = () => {
...
+    }
+
+    const message = await storage.createTapBackMessage(
+      connection.user.username,
+      tapInfo.tapResponse.chipIssuer
+    );
+    await sendMessages({
+      authToken: session.authTokenValue,
+      messages: [message],
+    });
```
Comment (stevenelleman) [inline apps/frontend/src/pages/people/[username].tsx:230]: try / catch to handle thrown error?

apps/frontend/src/pages/_app.tsx
```diff
@@ -29,6 +30,29 @@ export default function MyApp({ Component, pageProps }: AppProps) {
...
+      const user = await storage.getUser();
+      const session = await storage.getSession();
+      if (!user || !session) {
+        return;
+      }
+
+      const messages = await fetchMessages({
+        authToken: session.authTokenValue,
+        lastMessageFetchedAt: user.lastMessageFetchedAt,
+      });
```
stevenelleman, line 46: try / catch to handle thrown error?

apps/frontend/src/pages/people/[username].tsx
```diff
@@ -83,6 +87,17 @@ const CommentModal: React.FC<CommentModalProps> = ({
...
+          <div className="flex flex-col mt-2">
+            <AppButton
+              onClick={async () => {
+                setHasTappedBack(true);
+                await onTapBack();
+              }}
+              disabled={hasTappedBack}
+            >
+              Share socials back
+            </AppButton>
```
Comment (stevenelleman) [inline apps/frontend/src/pages/people/[username].tsx:99]: Hmm, I prefer that tapback is automatic so that the exchange is symmetric. It feels (1) like an equal relationship and (2) eliminates a step that might otherwise be skipped / forgotten. Just a personal preference, feel free to disregard.

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  const { user, session } = await getUserAndSession();
+
+  const connectionTapBacks: Record<string, TapBackMessage[]> = {};
+
+  let lastMessageTimestamp: Date | null = null;
+  for (const message of messages) {
+    // Ensure messages are in order and update lastMessageTimestamp
+    if (lastMessageTimestamp && message.createdAt < lastMessageTimestamp) {
+      throw new Error("Messages are not in order");
```
stevenelleman, line 32: I think this error would be surfaced to the user in `MyApp`.

apps/frontend/src/lib/message/index.ts
```diff
@@ -0,0 +1,168 @@
...
+  const response = await fetch(`${BASE_API_URL}/message`, {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify(request),
+  });
+
+  if (!response.ok) {
+    throw new Error(`Failed to send message: ${response.statusText}`);
```
stevenelleman, line 30: Unhandled runtime error.

apps/frontend/src/lib/message/index.ts
```diff
@@ -0,0 +1,168 @@
...
+    `${BASE_API_URL}/message?authToken=${encodeURIComponent(
+      authToken
+    )}${lastMessageFetchedAtParam}`,
+    {
+      method: "GET",
+    }
+  );
+
+  if (!response.ok) {
+    throw new Error(`Failed to fetch messages: ${response.statusText}`);
```
stevenelleman, line 53: Unhandled runtime error.

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
stevenelleman: If messages are out of order how can / should we resolve the issue? Would it be better to sort the messages before the loop and / or return an ordering warning?

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
```diff
@@ -0,0 +1,151 @@
...
+      } else {
+        connectionTapBacks[senderSignaturePublicKey] = [tapBackMessage];
+      }
+    } catch (error) {
+      console.error("Error processing message:", error);
+    }
+  }
+
+  if (!lastMessageTimestamp) {
+    throw new Error("No messages found");
```
stevenelleman, line 60: Unhandled runtime error.

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
```diff
@@ -0,0 +1,151 @@
...
+  const newBackupEntries: CreateBackupData[] = [];
+
+  Object.entries(connectionTapBacks).forEach(
+    ([senderSignaturePublicKey, tapBackMessages]) => {
+      if (tapBackMessages.length === 0) {
+        return;
+      }
+
+      const connectionUsername = tapBackMessages[0].user.username;
+      let updatedConnection: Connection = user.connections[connectionUsername];
```
stevenelleman, line 72: `Connection | null`?

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
```diff
@@ -0,0 +1,151 @@
...
+            tapBackMessage.user.psiPublicKeyLink
+        ) {
+          throw new Error("Connection cannot change username or keys");
+        }
+
+        updatedConnection = {
+          ...updatedConnection,
+          user: {
+            ...updatedConnection.user,
+            ...tapBackMessage.user,
```
stevenelleman, line 111: I'm not familiar with this syntax -- I'm assuming user is first set to values from updatedConnection and then tapBackMessage overrides non-matching values?

apps/frontend/src/lib/storage/types/user/connection/index.ts
```diff
@@ -1,19 +1,22 @@
...
 import { PSIDataSchema } from "./psiData";
-import { nullToUndefined } from "@types";
+import { SentMessageSchema } from "./sentMessage";
 
 export const ConnectionSchema = z.object({
   user: UserDataSchema,
   comment: nullToUndefined(CommentDataSchema),
   taps: z.array(TapDataSchema),
   psi: nullToUndefined(PSIDataSchema),
+  sentMessages: nullToUndefined(z.array(SentMessageSchema)).default([]),
```
stevenelleman, line 14: Just to make sure I understand -- there's no need for receivedMessages because the "tapper" already has the connection's information, and there's no other message types (yet) which require it?

stevenelleman:
Tested manually and validated that tapback works following a refresh.

stevenelleman:
When you tap someone a second time, the tapback is still given as an option. Could we keep track on the connection whether we've tapped back and hide the UI if we already have?

stevenelleman:
When you edit a chip you're given an option to tapback and it fails when you click it. Either the UI should be hidden in the edit view or I think it should be fully supported (which actually could be a nice feature -- gives you a chance to share if you forgot or changed your mind).

stevenelleman:
Messaging question -- could messaging be used to automatically refresh the other person's PSI results (ie update "Refresh" to the results)? Could be nice, otherwise the other party may not click on the results. Less important, but it could also be nice to have an associated activity / notification so that you know it's available.

stevenelleman:
Nit -- this isn't your responsibility but some of the back buttons don't go to the previous screen (probably need to use history object rather than a fixed path).

stevenelleman:
Confirmed that tapper still has their tap back activity after they log out and log back in. Also confirmed the same is true for the other party.

apps/backend/src/lib/controller/postgres/prisma/message/message.ts
AndrewCLu: empty array

apps/backend/src/routes/message/index.ts
AndrewCLu: not much server can do since it doesnt have visibility into the messages themselves, just sees the structure

AndrewCLu: updated 500

apps/frontend/src/pages/people/[username].tsx
Comment (AndrewCLu) [reply apps/frontend/src/pages/people/[username].tsx:230]: good catch

apps/frontend/src/pages/people/[username].tsx
Comment (AndrewCLu) [reply apps/frontend/src/pages/people/[username].tsx:99]: this is for consent, i.e. i dont want my data to ever be shared automatically. but in the future we should have this be a configurable option where someone can turn this toggle on in settings

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
AndrewCLu: good catch, sorted before loop

apps/frontend/src/lib/message/index.ts
AndrewCLu: fixed

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
AndrewCLu: this should never happen

AndrewCLu: added comment

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
AndrewCLu: yup

apps/frontend/src/lib/storage/types/user/connection/index.ts
AndrewCLu: yes. received messages always get parsed into the Connection struct, whether its user data or taps

AndrewCLu:
> When you tap someone a second time, the tapback is still given as an option. Could we keep track on the connection whether we've tapped back and hide the UI if we already have?

do you mean tapping someone's chip twice in a row. if so, i think you should be able to tap back again

AndrewCLu:
> Messaging question -- could messaging be used to automatically refresh the other person's PSI results (ie update "Refresh" to the results after the _second_ party presses "Discover")? Could be nice, otherwise the other party may not click on the results. Less important, but it could also be nice to have an associated activity / notification so that you know it's available.

I think this would be a websockets feature

apps/frontend/src/lib/message/tapBack.ts
```diff
@@ -0,0 +1,70 @@
...
+}: GenerateTapBackMessageArgs): Promise<{
+  serializedMessage: string;
+  messageTimestamp: Date;
+}> => {
+  const msgNonce = 0; // No counter for tap backs (4 bytes)
+  const generatedRandomBytes = randomBytes(28); // 28 random bytes
+  const message = getCounterMessage(
+    msgNonce,
+    generatedRandomBytes.toString("hex")
+  );
```
stevenelleman, line 42: Comprehension question -- will the message be the same for all tap backs? And if so, what is its role?

apps/frontend/src/pages/people/[username].tsx
```diff
@@ -200,6 +215,29 @@ const UserProfilePage: React.FC = () => {
...
+    );
+    try {
+      await sendMessages({
+        authToken: session.authTokenValue,
+        messages: [message],
+      });
+      toast.success("Tap back sent successfully!");
+    } catch (error) {
+      console.error("Error sending tap back:", error);
+      toast.error("Failed to send tap back");
```
Comment (stevenelleman) [inline apps/frontend/src/pages/people/[username].tsx:237]: Consider using the support toast here?

stevenelleman (approved):

PR #86, opened 2024-10-16
upgrade psi

AndrewCLu:
- [x] run psi on shared connections
- [x] save psi data after running and load when viewing connection profile
- [x] add activity feed item for running psi

### DIFF
--- a/apps/frontend/src/features/psi/InteractivePSI.tsx
+++ b/apps/frontend/src/features/psi/InteractivePSI.tsx
@@ -11,7 +11,7 @@ import { supabase } from "@/lib/realtime";
 import { Card } from "@/components/cards/Card";
 import { Icons } from "@/components/Icons";
 import { AppButton } from "@/components/ui/Button";
-import { Connection, UserData } from "@/lib/storage/types/user";
+import { Connection, PSIData, UserData } from "@/lib/storage/types/user";
 import { Tag } from "@/components/ui/Tag";
 import { logClientEvent } from "@/lib/frontend/metrics";
 import { EMOJI_MAPPING, INTERESTS_LIST } from "@/common/constants";
@@ -40,6 +40,8 @@ interface InteractivePSIProps {
   otherPsiPublicKeyLink: string;
   userData: UserData;
   connections: Record<string, Connection>;
+  existingPSIOverlap: PSIData | undefined;
+  savePSIOverlap: (overlap: PSIData) => Promise<void>;
 }
 
 const InteractivePSI: React.FC<InteractivePSIProps> = ({
@@ -50,6 +52,8 @@ const InteractivePSI: React.FC<InteractivePSIProps> = ({
   otherPsiPublicKeyLink,
   userData,
   connections,
+  existingPSIOverlap,
+  savePSIOverlap,
 }) => {
   const [psiState, setPsiState] = useState<PSIState>(PSIState.NOT_STARTED);
   const [broadcastEvent, setBroadcastEvent] = useState<any>();
@@ -73,10 +77,7 @@ const InteractivePSI: React.FC<InteractivePSIProps> = ({
   const [indexMapping, setIndexMapping] = useState<
     Record<number, string[]> | undefined
   >();
-  const [overlap, setOverlap] = useState<{
-    overlapUsers: string[];
-    overlapLannaDesiredConnections: string[];
-  }>();
+  const [overlap, setOverlap] = useState<PSIData>();
 
   useEffect(() => {
     if (wantsToInitiatePSI && otherUserWantsToInitiatePSI) {
@@ -94,6 +95,11 @@ const InteractivePSI: React.FC<InteractivePSIProps> = ({
   // set up channel for PSI
   // TODO: Load in previous overlap indices if available
   useEffect(() => {
+    if (existingPSIOverlap) {
+      setOverlap(existingPSIOverlap);
+      setPsiState(PSIState.COMPLETE);
+    }
+
     const channel = supabase.channel(getChannelName(), {
       config: {
         presence: { key: selfSigPK },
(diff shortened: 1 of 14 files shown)


stevenelleman:
Tested flow locally, able to do an initial PSI, make new connections, and update the PSI to include the new connections.

apps/frontend/src/lib/activity/index.ts
```diff
@@ -54,3 +54,15 @@ export const createTapActivity = (
     timestamp: new Date(),
   };
 };
+
+export const PSIActivityDataSchema = z.object({
+  connectionUsername: z.string(),
+});
+export type PSIActivityData = z.infer<typeof PSIActivityDataSchema>;
```
stevenelleman, line 61: Move to types?

apps/frontend/src/features/psi/InteractivePSI.tsx
```diff
@@ -360,6 +391,66 @@ const InteractivePSI: React.FC<InteractivePSIProps> = ({
...
+      <div className="flex flex-col gap-4">
+        {overlap?.sharedConnections && overlap.sharedConnections.length > 0 && (
+          <div>
+            <h3 className="font-medium text-bold text-primary text-sm font-sans mb-2">
+              Shared Connections
+            </h3>
+            <ul className="list-disc list-inside">
+              {overlap.sharedConnections.map((user, index) => (
+                <li key={index} className="text-tertiary text-sm">
+                  {user}
```
stevenelleman, line 407: Consider linking to their contact? As soon as profile pics are available it'd be great to use.

apps/frontend/src/features/psi/InteractivePSI.tsx
```diff
@@ -82,6 +94,11 @@ const InteractivePSI: React.FC<InteractivePSIProps> = ({
   // set up channel for PSI
   // TODO: Load in previous overlap indices if available
```
stevenelleman, line 95: Nit: remove

apps/frontend/src/lib/storage/types/user/connection/psiData.ts
```diff
@@ -0,0 +1,10 @@
+import { nullToUndefined } from "@types";
+import { z } from "zod";
+
+export const PSIDataSchema = z.object({
+  sharedConnections: z.array(z.string()),
+  sharedLannaDesiredConnections: z.array(z.string()),
```
stevenelleman, line 6: Connections is overloaded. Change to `sharedLannaInterests`?

apps/frontend/src/pages/people/[username].tsx
```diff
@@ -184,6 +184,24 @@ const UserProfilePage: React.FC = () => {
...
+
+    logClientEvent("user-profile-psi-overlap-saved", {});
+    const connectionUsername = connection.user.username;
+    await storage.updatePSI(connectionUsername, overlap);
+
+    // Update local state
+    const user = await storage.getUser();
+    const session = await storage.getSession();
+    if (!user || !session || !user.connections[connectionUsername]) {
+      throw new Error("User not found");
```
Comment (stevenelleman) [inline apps/frontend/src/pages/people/[username].tsx:198]: Error is a bit misleading.

apps/frontend/src/lib/psi/index.ts
```diff
@@ -22,20 +23,81 @@ export const generatePSIKeyPair = async () => {
...
+      const lannaConnections = [
+        "getHealthy",
+        "cowork",
+        "enjoyMeals",
+        "learnFrontierTopics",
+        "findCollaborators",
+        "goExploring",
+        "party",
+        "doMentalWorkouts",
+      ];
```
stevenelleman, line 91: Move out of for loop, no need to redefine every time? Nit: change to `lannaInterests`

apps/frontend/src/lib/psi/index.ts
```diff
@@ -22,20 +23,81 @@ export const generatePSIKeyPair = async () => {
...
+const COMMON_CONNECTIONS_BIT_VECTOR_SIZE = 9000;
+
+export const generateBitVectorFromUserData = (
+  userData: UserData,
+  connections: Record<string, Connection>
+): { bitVector: Uint32Array; indexMapping: Record<number, string[]> } => {
+  const bitVector = new Uint32Array(PSI_BIT_VECTOR_SIZE).fill(0);
+  const bitIndexToUsername: Record<number, string[]> = {};
+
+  // 0-8999 reserved for common connections
```
stevenelleman, line 36: So you're leaving 9000 places open to ensure there's enough space for connections. How do all these empty indices affect the computation? Is it efficient enough that it doesn't matter?

apps/frontend/src/lib/psi/index.ts
```diff
@@ -22,20 +23,81 @@ export const generatePSIKeyPair = async () => {
...
+  for (const connection of Object.values(connections)) {
+    const { username, signaturePublicKey } = connection.user;
+    const hash = sha256(signaturePublicKey);
+    const bigIntHash = BigInt("0x" + hash);
+    const index = Number(
+      bigIntHash % BigInt(COMMON_CONNECTIONS_BIT_VECTOR_SIZE)
+    );
+    bitVector[index] = 1;
```
stevenelleman, line 44: What prevents multiple connections from not overlapping on the same index? The large denominator (9000)?

stevenelleman (approved):

apps/frontend/src/lib/psi/index.ts
stevenelleman: When a user is created could they be allocated an incremented index that's used for their bitvector position? 

The middleman can already calculate the index in this scheme (because the sigPubKey is known to them, right? Or is it fully encrypted?).

AndrewCLu: yeah there's nothing here to prevent overlap. this bucketing method will lead to errors, which is just unlikely for now. but ux of increasing vector size does become worse

apps/frontend/src/lib/psi/index.ts
AndrewCLu: yeah, this is just based on previous experimentation, i think index size 100,000 is where begins to be a ux hit

apps/frontend/src/lib/activity/index.ts
AndrewCLu: right now all activity types and schemas are defined here, and are only used in context of activity helpers. but in the future could be worth a refactor

apps/frontend/src/lib/psi/index.ts
AndrewCLu: yup, fixed

apps/frontend/src/features/psi/InteractivePSI.tsx
AndrewCLu: added

PR #87, opened 2024-10-16
Leaderboard Logic

stevenelleman:
This provides all the piping from DB to frontend for user leaderboard position and the top 100 leaderboard positions. 

Tomorrow I'll do more thorough testing by populating my local DB with more LeaderboardEntry rows. [completed]

I'll also see if updating the pagination logic is easy enough, to assess if an infinite scroll leaderboard should be considered. That being said, does that leak too much info about users in the system. [after convo with Vivek decided to stick with 100] 

It is intentional that I did not link the page from the profile yet, I have a slight preference for the feature to be silently deployed, verified in production, and then a link added to make its existence known. 

Otherwise the only thing this PR is lacking are frontend components for displaying the leaderboard data which will be in its own PR.


apps/backend/src/lib/controller/chip/managed/leaderboard/index.ts
```diff
@@ -35,3 +62,69 @@ ManagedChipClient.prototype.UpdateLeaderboardEntry = async function (
...
+
+    2. Select username and calculated row number from entries
+    subquery2 = SELECT username, row_number() OVER() as "leaderboardPosition" FROM (subquery1) AS sorted_entries
+
+    3. Select row number based on username
+    query = SELECT row_number FROM (subquery2) AS filtered_entries WHERE "username"='pass2';
+
+    Working query:
+    SELECT "leaderboardPosition" FROM (SELECT username, row_number() OVER() as "leaderboardPosition" FROM (SELECT * FROM "LeaderboardEntry" WHERE "chipIssuer"='TESTING' ORDER BY "tapCount" DESC) AS sorted_entries) AS rowed_sorted_entries WHERE "username"='pass2';
+    */
```
stevenelleman, line 84: This query seems too complex. I'll keep mulling over ways to simplify it. If others have ideas I'm all ears. It does work tho.

apps/frontend/src/lib/chip/leaderboard.ts
```diff
@@ -19,7 +64,7 @@ export async function updateLeaderboardEntry(
 
   try {
     const response = await fetch(
-      `${BASE_API_URL}/user/update_leaderboard_entry`,
+      `${BASE_API_URL}/chip/update_leaderboard_entry`,
```
stevenelleman, line 67: NOTE: I changed the leaderboard routes from user to chip. All the frontend and backend leaderboard code already lives in `chip/` directories, not `user/`. This feels sensible to me, I think leaderboards should be "owned" by the chip issuer.

packages/types/src/chip/index.ts
```diff
@@ -79,3 +79,59 @@ export const ChipTapResponseSchema = z.object({
...
+
+export type GetLeaderboardEntryRequest = z.infer<
+  typeof GetLeaderboardEntryRequestSchema
+>;
+
+export const LeaderboardEntrySchema = z.object({
+  username: z.string(),
+  chipIssuer: z.string(),
+  tapCount: z.number(),
+  // In the future could imagine name, chip fields being added to allow custom leaderboards corresponding to specific chips
```
stevenelleman, line 106: Picturing a world where we have more leaderboards I imagine all we'd need to do is add a name field for the specific leaderboard the entry corresponds to.

packages/types/src/user/index.ts
```diff
@@ -193,11 +193,4 @@ export type VerifySigninTokenResponse = z.infer<
...
 >;
 
-export const UpdateLeaderboardEntryRequestSchema = z.object({
-  authToken: z.string(),
-  chipIssuer: ChipIssuerSchema,
-});
 
-export type UpdateLeaderboardEntryRequest = z.infer<
-  typeof UpdateLeaderboardEntryRequestSchema
->;
```
stevenelleman, line 203: Moved these to `chip/` types.

apps/backend/src/lib/controller/chip/managed/leaderboard/index.ts
```diff
@@ -1,6 +1,33 @@
...
+  ChipIssuer,
+  errorToString,
+  LeaderboardEntry,
+  LeaderboardEntrySchema,
+  LeaderboardPosition,
+  LeaderboardPositionSchema
+} from "@types";
 import { ManagedChipClient } from "../client";
 
+ManagedChipClient.prototype.GetLeaderboardEntry = async function (
```
stevenelleman, line 11: `GetLeaderboardEntry` may not be used by the frontend. If it isn't I'm inclined to keep it -- it was useful for testing.

apps/frontend/src/pages/leaderboard/[chipIssuer].tsx
```diff
@@ -0,0 +1,72 @@
...
+  "edge_lanna": ChipIssuer.EDGE_CITY_LANNA,
+  "devcon": ChipIssuer.DEVCON_2024,
+  "testing": ChipIssuer.TESTING,
+};
+
+const communitiesHumanReadable : { [key: string]: string } = {
+  "edge_lanna": "Edge City Lanna",
+  "devcon": "Dev Con 2014",
+  "testing": "Testing",
+};
```
Comment (stevenelleman) [inline apps/frontend/src/pages/leaderboard/[chipIssuer].tsx:23]: Should these be moved to constants or can they live here?

apps/frontend/src/pages/leaderboard/[chipIssuer].tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  if (!userLeaderboardPosition || !leaderboardEntries || !communityName) {
+    return (
+      <div className="flex min-h-screen justify-center items-center text-center">
+        <CursiveLogo isLoading />
+      </div>
+    );
+  }
+
+  return;
```
Comment (stevenelleman) [inline apps/frontend/src/pages/leaderboard/[chipIssuer].tsx:69]: This is populated in https://github.com/cursive-team/connections/pull/90

apps/backend/src/lib/controller/chip/managed/leaderboard/index.ts
```diff
@@ -35,3 +62,69 @@ ManagedChipClient.prototype.UpdateLeaderboardEntry = async function (
...
+  try {
+    // In this scheme there's currently no way to decide ties, ie ties may change order.
+    const topEntries = await this.prismaClient.leaderboardEntry.findMany({
+      where: {
+        chipIssuer,
+      },
+      take: count,
+      orderBy: {
+        tapCount: "desc",
+      },
```
stevenelleman, line 117: TODO: order on username as well? That way the order of ties would remain consistent.

apps/backend/src/routes/chip/leaderboard_entry.ts
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * @route POST /api/chip/update_leaderboard_entry
+ * @desc Updates the leaderboard entry for a user
+ */
+router.post(
+  "/update_leaderboard_entry",
+  async (
+    req: Request<{}, {}, UpdateLeaderboardEntryRequest>,
+    res: Response<{} | ErrorResponse>
```
stevenelleman, line 62: Is `{}` equivalent to `any` in this context?

RiverRuby: it's equal to returning nothing in the response

apps/backend/src/lib/controller/chip/managed/leaderboard/index.ts
RiverRuby: LGTM! Although I would prefer to convert this into pure prisma queries, so that it's easier to read and maintain. I'll add an issue for that in v1.3 milestone

RiverRuby: Made it #100

apps/frontend/src/pages/leaderboard/[chipIssuer].tsx
Comment (RiverRuby) [reply apps/frontend/src/pages/leaderboard/[chipIssuer].tsx:23]: we can keep them here until they start getting used in many more places

apps/frontend/src/lib/chip/leaderboard.ts
RiverRuby: yep this was a great choice, appreciate the thought there

RiverRuby (approved):
LGTM! Really great work here. Code is clean and organized and was easy to get through

apps/backend/src/lib/controller/chip/managed/leaderboard/index.ts
stevenelleman: Sounds good, definitely agreed, when possible I tried to default to prisma, but found it difficult to use for a few of these queries.

PR #90, opened 2024-10-17
Leaderboard UI 

stevenelleman:
This is approximately 95% of the [Cursive Connections leaderboard design](https://www.figma.com/design/wT80Rr4FeVdCJdfRw4JEV8/Cursive-Connections?node-id=4200-3450&node-type=canvas&m=dev). 

Design vs PR: 

Screenshot 2024-10-18 at 2 15 05 AM

image

*Note: the leaderboard positions are repeated because I was ensuring ties work. In the case of ties, the first rank + styling is shared between all tied users. Consistent order of tied users is not guaranteed. 

Here are the discrepancies: (Note: 1-4 are labeled on the screenshot)

1. The padding isn't correct. I wasn't able to reduce it. [removed a class, got a bit closer, still a tiny bit off] 
2. I wasn't able to figure out how to `text-align: right` the `Total taps`. [Solution: `display: content`, but this now center aligns the username] 
3. The gradient is incorrect. The gradient values from the designs were `#7A74BC, #FF9DF8, #FB5D42, #FF0000` but when applied to `linear-gradient()` they don't look right. [figured out the right gradient] 
4. I did not include "them" because I didn't understand what it was doing. [Answered in standup, going to include this feature when I add the user profile link] 
5. I intentionally did not include the profile in this PR. I would prefer to deploy this, verify it works in prod, and then make the feature discoverable by adding the profile link. 

Open question: 
- Should the design really not have the bottom buttons (profile, activities, connections)? [Answered in standup, fine for nav footer to not be included] 

https://github.com/user-attachments/assets/d455012e-3ebe-4a30-8a12-c730fcb2a1cb

### DIFF
--- a/apps/frontend/src/layouts/AppLayout.tsx
+++ b/apps/frontend/src/layouts/AppLayout.tsx
@@ -36,7 +36,11 @@ export default function AppLayout({
   return (
     <div>
       {back && (
-        <div className="sticky top-0 h-12 flex items-center border-b border-b-quaternary/20 bg-white z-20">
+        <div className={cn(
+          "sticky top-0 h-12 flex items-center bg-white z-20",
+          headerDivider && "border-b border-b-quaternary/20"
+        )}
+        >
           <div className="px-4">
             <Link className="flex gap-1 items-center" href={back?.href ?? "/"}>
               <BackIcon size={12} />
--- a/apps/frontend/src/layouts/AppLayout.tsx
+++ b/apps/frontend/src/layouts/AppLayout.tsx
@@ -36,11 +36,7 @@ export default function AppLayout({
   return (
     <div>
       {back && (
-        <div className={cn(
-          "sticky top-0 h-12 flex items-center bg-white z-20",
-          headerDivider && "border-b border-b-quaternary/20"
-        )}
-        >
+        <div className="sticky top-0 h-12 flex items-center border-b border-b-quaternary/20 bg-white z-20">
           <div className="px-4">
             <Link className="flex gap-1 items-center" href={back?.href ?? "/"}>
               <BackIcon size={12} />
--- a/apps/frontend/src/pages/leaderboard/[chipIssuer].tsx
+++ b/apps/frontend/src/pages/leaderboard/[chipIssuer].tsx
@@ -159,103 +159,64 @@ const LeaderboardPage: React.FC = () => {
           // Update lastTapCount
           lastTapCount = entry.tapCount
 
-          if (position == 1) {
-            return (
-              <div className="h-6 px-4 justify-between items-center inline-flex" style={{width: "100%", marginBottom: "4px", marginTop: "4px"} as CSSProperties}>
-                <div className="grow shrink basis-0 h-6 justify-start items-center gap-3 flex">
-                  <div
-                    className="w-10 h-6 px-0.5 py-2 bg-[#090909] rounded-[67px] justify-center items-center gap-2 flex">
-                    <div className="text-center text-white text-sm font-medium font-['DM Sans'] leading-tight">1</div>
-                  </div>
-                  <div
-                    className="grow shrink basis-0 text-[#090909] text-base font-medium font-['DM Sans'] leading-snug">{entry.username}
-                  </div>
-                  <div className="justify-start items-start gap-[5px] flex">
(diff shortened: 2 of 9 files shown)


apps/frontend/src/pages/leaderboard/[chipIssuer].tsx
```diff
@@ -66,7 +68,138 @@ const LeaderboardPage: React.FC = () => {
...
+            styling.positionColor = "bg-[#090909]";
+            styling.positionTextColor = "text-white"; // Done
+            styling.fontStyling = "text-[#090909] font-medium";
+          }
+
+          if (entry.username == userLeaderboardPosition.username) {
+            styling.positionColor = "bg-[#f74227]";
+            styling.positionTextColor = "text-white"; // Done
+            styling.fontStyling = "text-[#090909] font-medium";
+            username += " (me)"; // Done
```
Comment (stevenelleman) [inline apps/frontend/src/pages/leaderboard/[chipIssuer].tsx:163]: Will remove these comments.

apps/frontend/src/layouts/AppLayout.tsx
```diff
@@ -45,7 +45,11 @@ export default function AppLayout({
   return (
     <OnlyMobileLayout>
       {back && (
-        <div className="sticky top-0 h-12 flex items-center border-b border-b-quaternary/20 bg-white z-20">
+        <div className={cn(
+          "sticky top-0 h-12 flex items-center bg-white z-20",
+          headerDivider && "border-b border-b-quaternary/20"
+        )}
+        >
```
stevenelleman, line 52: This allows the divider to be removed from the top header (the header with the back button).

stevenelleman:
TODO: in the video there's an example of a long name overflowing off the screen. I need to update this so that name fades out (while keeping the tap count in the same position) instead.

apps/frontend/src/pages/leaderboard/[chipIssuer].tsx
```diff
@@ -66,7 +68,138 @@ const LeaderboardPage: React.FC = () => {
...
+            <div key={index}>
+            <div className="h-6 px-4 justify-between items-center inline-flex"
+                 style={{width: "100%", marginBottom: "4px", marginTop: "4px"} as CSSProperties}>
+              <div className="grow shrink basis-0 h-6 justify-start items-center gap-3 flex">
+                <div className={`w-10 h-6 px-0.5 py-2 ${styling.positionColor} rounded-[67px] justify-center items-center gap-2 flex`}>
+                  <div
+                    className={`text-center ${styling.positionTextColor} text-[#090909] text-sm font-medium font-['DM Sans'] leading-tight`}>{position}</div>
+                </div>
+                <div
+                  className={`grow shrink basis-0 ${styling.fontStyling} font-['DM Sans'] leading-tight`}>{username}
```
Comment (stevenelleman) [inline apps/frontend/src/pages/leaderboard/[chipIssuer].tsx:185]: I'd like to figure out how to apply `overflow: hidden` on this attribute but I need to define a max width somehow, I'm not sure how to set a max width that's appropriate for multiple phone screen types. This issue only applies to really long usernames, so it's not first priority.

apps/frontend/src/pages/leaderboard/[chipIssuer].tsx
```diff
@@ -37,6 +29,12 @@ const LeaderboardPage: React.FC = () => {
...
     const fetchInfo = async () => {
       if (typeof chipIssuer == "string") {
+        if (!Object.keys(communitiesEnum).includes(chipIssuer)) {
+          toast.error("Invalid chip issuer.");
+          router.push("/");
+          return;
+        }
+
         const chipIssuerMapping = communitiesEnum[chipIssuer];
         const chipIssuerEnum = ChipIssuerSchema.parse(chipIssuerMapping);
```
Comment (RiverRuby) [inline apps/frontend/src/pages/leaderboard/[chipIssuer].tsx:40]: There's a bug where if a user hasn't tapped anyone else, this leaderboard query errors out.

RiverRuby (approved):
LGTM! fixed frontends to match the mocks, and fixed a bug where if you're not on leaderboard yet (aka just created account) it errors out

apps/frontend/src/pages/leaderboard/[chipIssuer].tsx
Comment (stevenelleman) [reply apps/frontend/src/pages/leaderboard/[chipIssuer].tsx:40]: Ah nice find, will address right now.

PR #107, opened 2024-10-20
Import Data Using OAuth

stevenelleman:
This PR uses the oauth access token to import Strava monthly run data and converts it to a LeaderboardEntry. 

NOTE: this should fail ATM, I _did not_ push corresponding LeaderboardEntry changes (schema, types) because @AndrewCLu is owning those changes. This PR should be rebased onto his work and then adjusted as needed.


apps/frontend/src/pages/oauth/exchange_token.tsx
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        if (accessToken) {
+          const mapping = OAUTH_APP_MAPPING[stateStr];
+          if (mapping) {
+            // TODO: Need better (1) chipIssuer selection when more communities are added, (2) Data Option selection when more import options are available
+            const leaderboardEntry = await importOAuthData(user.userData.username, ChipIssuer.EDGE_CITY_LANNA, accessToken, mapping.data_options[0]);
+
+            // TODO: Insert LeaderboardEntry into DB
+            console.log("Leaderboard Entry", leaderboardEntry);
```
stevenelleman, line 45: All you need to do here is to create / update the Entry here.

PR #114, opened 2024-10-22
edit private notes and emojis

AndrewCLu:


apps/frontend/src/pages/people/[username].tsx
```diff
@@ -473,27 +524,6 @@ const UserProfilePage: React.FC = () => {
...
-
-          {connection?.comment?.emoji && (
-            <div className="flex flex-col gap-2 py-4 px-4">
-              <span className="text-sm font-semibold text-primary font-sans">
-                Your Label
-              </span>
-
-              <p className="text-2xl mt-2">{connection?.comment?.emoji}</p>
-            </div>
-          )}
```
Comment (stevenelleman) [inline apps/frontend/src/pages/people/[username].tsx:496]: Removed to reduce the chances the info from being leaked accidentally?

apps/frontend/src/pages/people/[username].tsx
```diff
@@ -186,7 +216,8 @@ const UserProfilePage: React.FC = () => {
       }
 
       setConnection(user.connections[connection.user.username]);
-      handleCloseTapModal();
+      setShowCommentModal(false);
+      await storage.deleteSavedTapInfo();
```
Comment (stevenelleman) [inline apps/frontend/src/pages/people/[username].tsx:220]: Why is the tap info being deleted?

stevenelleman (approved):
Tested locally and works.

apps/frontend/src/pages/people/[username].tsx
Comment (AndrewCLu) [reply apps/frontend/src/pages/people/[username].tsx:220]: this is just saved tap info, its the temporary tap thats saved from when you tap so the info is surfaced for registration, etc.

apps/frontend/src/pages/people/[username].tsx
Comment (AndrewCLu) [reply apps/frontend/src/pages/people/[username].tsx:496]: yeah exactly

PR #124, opened 2024-10-24
Janine Request: Gym Emoji 

stevenelleman:
In two HIIT classes Janine has requested an emoji to indicate people she can work out with. I replaced the halloween icon because it felt the disco ball fulfilled the same role. 

I originally tried to find a weight to avoid the entire matter of ethnicity but none were available. If I find anything I'll swap it out.


stevenelleman:
One way to avoid changing previous behavior is to simply extend emoji options. Would this be useful for devcon? I feel like more options could be useful.

RiverRuby:
Closing because not relevant anymore

PR #130, opened 2024-10-25
save oauth token and application data to backup

AndrewCLu:


apps/frontend/src/lib/storage/index.ts
```diff
@@ -25,6 +25,8 @@ export interface ClientStorage {
   getUserAndSession(): Promise<{ user: User; session: Session }>;
   getUser(): Promise<User | undefined>;
   getSession(): Promise<Session | undefined>;
+  saveOAuthAccessToken(app: string, token: AccessToken): Promise<void>;
+  getOAuthAccessToken(app: string): Promise<AccessToken | undefined>;
```
stevenelleman, line 29: Whoops, good catch.

apps/frontend/src/lib/storage/localStorage/user/oauth.ts
```diff
@@ -0,0 +1,38 @@
...
+  token: AccessToken
+): Promise<void> => {
+  const { user, session } = await getUserAndSession();
+
+  const oauthBackup = createOAuthBackup({
+    email: user.email,
+    password: session.backupMasterPassword,
+    oauthData: {
+      app,
+      token,
```
stevenelleman, line 17: Just to check -- it's possible to update this in the future for scope? Or if the scope changes should we just overwrite the existing token?

AndrewCLu: yeah we can just overwrite

stevenelleman (approved):

PR #136, opened 2024-10-27
Delete Revoked Token If Import Fails

stevenelleman:
I wanted to make a screen recording of the full GH authorization flow. I revoked my current oauth token in GH, thinking we had handled this case. When I attempted to import GH data the flow failed -- the access token had been revoked but looked valid to the client so it kept using the same token rather than removing it. This PR removes the access token if an import fails, forcing the client to create a new access token.

### DIFF
--- a/apps/frontend/src/lib/storage/localStorage/user/oauth.ts
+++ b/apps/frontend/src/lib/storage/localStorage/user/oauth.ts
@@ -36,3 +36,15 @@ export const getOAuthAccessToken = async (
 
   return undefined;
 };
+
+export const deleteOAuthAccessToken = async (
+  app: string
+): Promise<void> => {
+  const { user } = await getUserAndSession();
+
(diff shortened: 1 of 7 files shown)


apps/frontend/src/lib/storage/localStorage/user/oauth.ts
```diff
@@ -36,3 +36,15 @@ export const getOAuthAccessToken = async (
 
   return undefined;
 };
+
+export const deleteOAuthAccessToken = async (
```
AndrewCLu, line 40: this doesnt actually delete the oauth token from the backup. to do that you probably want to create a new backup type called DELETE_OAUTH_ACCESS_TOKEN or something like that

AndrewCLu (commented):
is the goal to delete oauth from local storage and backup? if so, need to update logic

AndrewCLu (approved):
lgtm pending testing

PR #139, opened 2024-10-27
implement location taps

AndrewCLu:


RiverRuby (approved):
i think it makes sense. i don't love that the chip type has so many fields, but each locationChip should also have an owner so i think it makes sense to combine them

apps/backend/prisma/schema.prisma
```diff
@@ -84,6 +84,11 @@ model Chip {
   ownerEncryptionPublicKey String?
   ownerPsiPublicKeyLink    String?
   ownerUserData            Json? // Nullable JSONB field
+  isLocationChip           Boolean?
+  locationId               String?   @unique
+  locationName             String?
+  locationDescription      String?
+  locationData             Json? // Nullable JSONB field
```
stevenelleman, line 91: Who's the owner of a location chip? Is it the admin that set it up? 

I think for this activation this makes sense, but for a future one these fields are basically repeated in the connection chip. Could be interesting to consider what a generic chip would look like for future activations.

AndrewCLu: I think the owner could be the person who owns the chip but the issuer is the community that issues the chip, and these theoretically could be different

apps/backend/prisma/seed.ts
```diff
@@ -22,7 +23,35 @@ async function main() {
...
+      const { signingKey, verifyingKey } = generateSignatureKeyPair();
+      return {
+        chipIssuer: "TESTING",
+        chipId: chip.chipId,
+        chipVariant: "NTAG212",
+        chipIsRegistered: true,
+        chipPublicKey: verifyingKey,
+        chipPrivateKey: signingKey,
+        chipTapCount: 0,
+        isLocationChip: true,
```
stevenelleman, line 45: could be good to pass in an option for seeding location vs user chip

apps/backend/src/lib/controller/chip/managed/tap/index.ts
```diff
@@ -10,6 +10,7 @@ ManagedChipClient.prototype.GetTapFromChip = async function (
       this.prismaClient,
       tapParams
     );
+    console.log(tapResponse);
```
stevenelleman, line 13: nit

apps/frontend/src/pages/locations/[id].tsx
```diff
@@ -84,9 +99,83 @@ const LocationTapModal: React.FC<LocationTapModalProps> = ({
...
+
+      // Compute the days of the week that the user has tapped in
+      const tapDays = computeWeeklyTapDays(
+        location.taps?.map((tap) => new Date(tap.timestamp))
+      );
+      setWeeklyTapDays(tapDays);
+
+      const savedTapInfo = await storage.loadSavedTapInfo();
+      // Delete saved tap info after fetching
+      await storage.deleteSavedTapInfo();
```
Comment (stevenelleman) [inline apps/frontend/src/pages/locations/[id].tsx:134]: Delete within the if case? Should we still delete the tap info if it's not equal to the id?

stevenelleman (approved):

apps/backend/src/lib/controller/chip/managed/tap/index.ts
AndrewCLu: thx

apps/frontend/src/lib/storage/localStorage/user/location/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  const previousLocation = user.locations?.[tap.locationId];
+
+  if (
+    previousLocation &&
+    previousLocation.chipPublicKey &&
+    previousLocation.chipPublicKey !== tap.chipPublicKey
+  ) {
+    throw new Error("Cannot update location with a different chip");
```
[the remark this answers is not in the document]
AndrewCLu: yup agreed, we need to add authentication for this

apps/frontend/src/pages/locations/[id].tsx
Comment (AndrewCLu) [reply apps/frontend/src/pages/locations/[id].tsx:134]: yes i think we should, it shouldnt happen but i think case it does i think it makes sense to delete it

PR #143, opened 2024-10-29
Websocket Messaging

stevenelleman:


apps/frontend/src/pages/people/[username].tsx
```diff
@@ -1,27 +1,35 @@
...
-import { SupportToast } from "@/components/ui/SupportToast";
-import { ERROR_SUPPORT_CONTACT } from "@/constants";
-import { sendMessages } from "@/lib/message";
+import {TensionSlider} from "../tensions";
+import {Icons} from "@/components/Icons";
+import {SupportToast} from "@/components/ui/SupportToast";
+import {ERROR_SUPPORT_CONTACT} from "@/constants";
+import {sendMessages} from "@/lib/message";
+import {wsRequest} from "@/lib/ws";
+import {getConnectionPubKey, getUserSigPubKey} from "@/lib/user";
```
Comment (stevenelleman) [inline apps/frontend/src/pages/people/[username].tsx:32]: I'll change this back.

apps/backend/src/index.ts
```diff
@@ -6,27 +6,128 @@ import healthRoutes from "./routes/health";
...
+
+    try {
+      switch (req.type) {
+        case WebSocketRequestTypes.CONNECT:
+          if (!req.senderSigPubKey) {
+            return handleError(socket, req.senderSigPubKey, "Missing target.")
+          }
+
+          // Set record for future lookup
+          clients[req.senderSigPubKey] = socket;
```
stevenelleman, line 99: Currently I use the signing public key as the identifier for a websocket. While this seemed like the simplest option, a more human-readable / friendly option would be to use username (which I believe is unique).

apps/frontend/src/lib/ws/index.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    switch (resp.type) {
+      case WebSocketResponseTypes.MSG:
+        const createMessage: CreateMessageData = CreateMessageDataSchema.parse(resp.payload);
+
+        // Convert create to message by adding created at
+        const message: MessageData = MapCreateMessageDataToMessageData(createMessage)
+
+        if (message) {
+          await storage.processNewMessages([message]);
```
stevenelleman, line 61: I need to do a check in the refresh message to avoid double-saving the same messages.

apps/backend/src/ws.ts
```diff
@@ -0,0 +1,101 @@
...
+  errorToString,
+  MapRequestToResponse,
+  WebSocketErrorPayload,
+  WebSocketRequest,
+  WebSocketRequestTypes,
+  WebSocketResponse,
+  WebSocketResponseTypes
+} from "@types";
+
+const WS_PORT = process.env.WS_PORT || 8090;
```
stevenelleman, line 14: NOTE: new WS_PORT env variable.

apps/frontend/src/lib/storage/localStorage/utils.ts
```diff
@@ -8,6 +8,7 @@ export const saveToLocalStorage = (key: string, value: string): void => {
   localStorage.setItem(key, value);
 };
 
+// uncaughtException: ReferenceError: localStorage is not defined
 export const getFromLocalStorage = (key: string): string | null => {
```
stevenelleman, line 12: Been running into this runtime issue where the browser is confusing the native window LocalStorage with the node LocalStorage.

apps/frontend/src/pages/people/index.tsx
```diff
@@ -30,7 +30,7 @@ const PeoplePage: React.FC = () => {
     };
 
     fetchConnections();
-  }, []);
+  }, [connections]);
```
stevenelleman, line 33: NOTE: I was having issues figuring out how to refresh the frontend component after the WS message is received and processed. As the value that's changing is connections, making `useEffect` dependent on the value leads to the desired refresh behavior. I'm open to other refresh approaches, but this one worked.

packages/types/src/ws/index.ts
```diff
@@ -0,0 +1,53 @@
+import {z} from "zod";
+
+export enum WebSocketRequestTypes {
+  CONNECT = "connect",
+  CLOSE = "close",
+  MSG = "message",
+}
+
+export enum WebSocketResponseTypes {
+  // CONNECT = "connect", // NOTE: we may need a server response to confirm connection / client info has been successfully received
```
stevenelleman, line 10: Thoughts?

apps/frontend/src/config.ts
```diff
@@ -11,6 +11,8 @@ export const FRONTEND_URL =
 export const BASE_API_URL =
   process.env.NEXT_PUBLIC_API_URL || "http://localhost:8080/api";
 
+export const BASE_API_WS = process.env.NEXT_PUBLIC_API_WS || "ws://localhost:8090/socket"
```
stevenelleman, line 14: NOTE: New frontend BASE_API_WS env var.

apps/frontend/src/lib/ws/index.ts
```diff
@@ -0,0 +1,100 @@
...
+      // WS connection is established earlier but need user public key for any requests
+      wsConnectUser(session.authTokenValue, sigPubKey);
+    }
+  } else {
+    console.warn("No user public signing key available, unable to establish websocket connection")
+  }
+};
+
+// @ts-expect-error : solve "TS2702: WebSocket only refers to a type, but is being used as a namespace here."
+wsClient.onmessage = async (e: WebSocket.MessageEvent) => {
```
stevenelleman, line 40: Haven't figured out how to fix this yet. I want to use the native WebSocket library to specify the event type.

apps/frontend/src/lib/storage/localStorage/utils.ts
stevenelleman: I don't see what change I introduced that could create this behavior. I imported ws and ws types, nothing to do with LocalStorage.

apps/frontend/src/lib/ws/index.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (sigPubKey && session) {
+      // WS connection is established earlier but need user public key for any requests
+      wsConnectUser(session.authTokenValue, sigPubKey);
+    }
+  } else {
+    console.warn("No user public signing key available, unable to establish websocket connection")
+  }
+};
+
+// @ts-expect-error : solve "TS2702: WebSocket only refers to a type, but is being used as a namespace here."
```
stevenelleman, line 40: Haven't figured out how to resolve this yet. I want to use native WebSockets and was having trouble importing the type.

apps/backend/src/index.ts
```diff
@@ -31,7 +33,11 @@ app.use("/", healthRoutes);
...
 const PORT = process.env.PORT || 8080;
-app.listen(PORT, () => {
+server.listen(PORT, () => {
   console.log(`Server is running on port ${PORT}`);
-});
\ No newline at end of file
+});
+
+// Use same server for websocket connection
+export const wsServer = new WebSocketServer({ server })
```
stevenelleman, line 43: I thought the vercel issue may relate to http and ws not sharing the same port. This change allows ws to piggyback on the same server as http. Alas, it did not solve the vercel issue :-(

stevenelleman: @AndrewCLu -- another hypothesis is that the backend server hasn't been deployed in render, so it's missing the websocket server.

apps/frontend/src/lib/storage/localStorage/utils.ts
stevenelleman: Even when I use `window.localStorage.getItem(key);` it complains with the same error even though I can follow the reference correctly, very weird.

apps/frontend/package.json
```diff
@@ -39,7 +39,6 @@
     "zod": "^3.23.8"
   },
   "devDependencies": {
-    "@types/node": "^20",
```
stevenelleman, line 42: I may add this back, I was trying to see if the node import was causing the conflict with localStorage (there's a ts localStorage and a node localStorage)

apps/frontend/src/lib/ws/index.ts
stevenelleman: Ah "use client"

stevenelleman:
Rather than using preview environment, instead should fail gracefully in both directions.

stevenelleman:
Attempted to test the the web sockets fail gracefully in both directions -- I changed the client to point to the wrong port and ensured that any error are handled and I refactored the client and server code to both be passive try / catch statements to maximize surface area of error handling.

apps/backend/src/index.ts
stevenelleman: I'm pretty sure this is the issue, resolving.

PR #147, opened 2024-10-29
add lanna workouts leaderboard

AndrewCLu:


apps/frontend/src/lib/chip/leaderboard.ts
```diff
@@ -42,6 +42,59 @@ export async function updateLeaderboardEntry(
   }
 }
 
+export async function updateWorkoutLeaderboardEntry(): Promise<void> {
```
stevenelleman, line 45: Could naming of this feature be changed to "WorkoutCheckIns"? I may also have a workout feature of some sort from more Strava data so it'd be nice to have more specific / differentiated naming.

apps/frontend/src/pages/community/index.tsx
```diff
@@ -121,14 +121,6 @@ export default function CommunityPage() {
...
           LeaderboardEntryType.TOTAL_TAP_COUNT
         );
-        weeklyDetails = await getUserLeaderboardDetails(
-          communityIssuer,
-          LeaderboardEntryType.WEEK_OCT_20_TAP_COUNT
-        );
-        weeklyEntries = await getTopLeaderboardEntries(
-          communityIssuer,
-          LeaderboardEntryType.WEEK_OCT_20_TAP_COUNT
-        );
```
stevenelleman, line 131: Rather than deleting old dashboards I wish they could be hidden somehow. I'm a sucker for digging into historical patterns / trends.

stevenelleman (approved):

apps/frontend/src/pages/community/index.tsx
AndrewCLu: yup! included in vivek's pr

apps/frontend/src/lib/chip/leaderboard.ts
AndrewCLu: this is under /chip, so would only refer to chip based leaderboard updates

PR #148, opened 2024-10-29
remove interactive psi, supabase, vercel blob

AndrewCLu:


stevenelleman:
LGTM:

Tested: 
- registration with passkey 
- registration with password 
- tapping between newly created accounts
- tapping between new and old accounts 
- tapbacks (after 2 refreshes) 
- edit profile 
- no errors seen across page views

stevenelleman (approved):

PR #150, opened 2024-10-30
halloween theme

kalidiagne:
To enable the Halloween theme, set the environment variable `NEXT_PUBLIC_HALLOWEEN_THEME="true"`

For customizing specific components, you can use the following example:
```
const hasHalloweenTheme = process.env.NEXT_PUBLIC_HALLOWEEN_THEME === "true";
```

This will allow you to apply a custom class based on the theme status. For general settings, refer to `settings.css`, where the variables are defined. Ensure that any changes made to a variable do not interfere with other components


stevenelleman:
Overall looks really good.

stevenelleman:
image

The contrast it a bit harsh, might be advisable to dim the profile components a bit.

apps/frontend/src/components/modals/SpookyModal.tsx
```diff
@@ -34,75 +34,77 @@ export interface ModalProps
...
+    label: "stick together at the party"
+  },
+  chillAndVibe: {
+    emoji: "😎",
+    label: "chill and vibe"
   },
+  introverse: {
+    emoji: "🎴",
+    label: "play the Introverse card game"
+  }
```
stevenelleman, line 87: image

Updated to be in line with the current design in freeze V1.2.

apps/frontend/src/components/modals/AstrologyModal.tsx
```diff
@@ -402,6 +402,7 @@ const AstrologyModal = ({
     if (step === steps?.length - 1) {
       onClose?.();
       setIsOpen(false);
+      setStep(0);
```
stevenelleman, line 405: Note: after completing, the step will be reset to 0 so that you'll edit from the start. This seemed more intuitive than editing in reverse order.

AndrewCLu (approved):

Issue #94, opened 2024-10-18
Make sure editing chip details is authorized


RiverRuby:
Full security requires signing updates.

PR #154, opened 2024-10-30
Basic Telegram notification bot

RiverRuby:
All code for a basic telegram notification bot are in this PR. The prod one is "CurtisConnectionBot", the environment variables for him are already upstreamed to prod Render.

To test you should make your own bot as Telegram only allows one running instance of a Bot at a time. Instructions for this are in README.md.

There's an example button to connect to the bot on `profile/index.tsx`. This can be removed until we have better designs for it, but I left it there for easy testing.

There's an example usage of the notification message when a user updates their chip profile. This can be removed once it is added in other places.

There is a migration in this PR, so we need to ensure that gets upstreamed to the backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,14 @@ Backend:
 - Run `pnpm prisma generate` to generate the prisma client
 - `pnpm run dev`
 
+Telegram:
+
+- Create a bot on Telegram using BotFather to test with
+- In `apps/backend/.env`:
+  - Set `TELEGRAM_BOT_API` to the bot api key
+  - Set `TELEGRAM_BOT_USERNAME` to be the bot username
+  - Set `TELEGRAM_BOT_START_DELAY_MS` to the delay in ms before starting the bot, defaults to 0, mostly relevant in prod
+
 Testing:
 
 - This is meant to be a mobile browser app. You can either test this on mobile browser by connecting it to localhost or using ngrok/similar service, or test from a desktop web browser using mobile view
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Backend:
 
 Telegram:
 
-- Create a bot on Telegram using BotFather to test with
+- Create a bot on Telegram using BotFather (https://t.me/BotFather) to test with
 - In `apps/backend/.env`:
   - Set `TELEGRAM_BOT_API` to the bot api key
   - Set `TELEGRAM_BOT_USERNAME` to be the bot username
--- a/apps/backend/src/lib/controller/chip/managed/client.ts
+++ b/apps/backend/src/lib/controller/chip/managed/client.ts
@@ -9,6 +9,7 @@ import {
   LeaderboardEntryType,
 } from "@types";
 import { Chip } from "@/lib/controller/chip/types";
+import { TelegramNotificationClient } from "../../notification/telegram/client";
 
 // NOTE: Hoist all prototype methods -- if you do not import the method file, the method(s) will evaluate to undefined at runtime
 import("@/lib/controller/chip/managed/update");
@@ -18,9 +19,11 @@ import("@/lib/controller/chip/managed/tap");
 
 export class ManagedChipClient implements iChipClient {
   prismaClient: PrismaClient;
+  notificationClient: TelegramNotificationClient;
 
   constructor() {
     this.prismaClient = new PrismaClient();
+    this.notificationClient = new TelegramNotificationClient();
   }
 
   // @ts-expect-error (ts2391: function implementation does not immediately follow declaration)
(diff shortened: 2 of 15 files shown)


RiverRuby:
Oh also, I realized updateChip endpoint wasn't protected, so this closes #94

stevenelleman:
Confirmed that I receive a notification when my chip is updated.

README.md
```diff
@@ -28,6 +28,14 @@ Backend:
 - Run `pnpm prisma generate` to generate the prisma client
 - `pnpm run dev`
 
+Telegram:
+
+- Create a bot on Telegram using BotFather to test with
```
stevenelleman, line 33: Include link: https://t.me/BotFather

apps/backend/src/lib/controller/chip/managed/client.ts
```diff
@@ -18,9 +19,11 @@ import("@/lib/controller/chip/managed/tap");
 
 export class ManagedChipClient implements iChipClient {
   prismaClient: PrismaClient;
+  notificationClient: TelegramNotificationClient;
```
stevenelleman, line 22: This seems redundant as the client is already included in the Controller. You can remove this if you move the notification here: https://github.com/cursive-team/connections/blob/main/apps/backend/src/lib/controller/index.ts#L141

apps/backend/src/lib/controller/notification/telegram/client.ts
```diff
@@ -0,0 +1,114 @@
+import { Bot } from "grammy";
+import { iNotificationClient } from "../interfaces";
+import { PrismaClient } from "@prisma/client";
+
+export class TelegramNotificationClient implements iNotificationClient {
+  private bot: Bot | undefined;
+  private readonly maxRetries = 3;
+  private readonly retryDelay = 5000;
+  prismaClient: PrismaClient | undefined;
```
stevenelleman, line 9: Should it ever be undefined? If not, remove?

stevenelleman: By initializing prismaClient before the env var check you remove `undefined`. Will push.

apps/backend/src/lib/controller/notification/telegram/client.ts
```diff
@@ -0,0 +1,114 @@
...
+    for (let i = 0; i < this.maxRetries; i++) {
+      try {
+        await this.setupBot();
+        console.log("Telegram notification client initialized");
+        return;
+      } catch (error) {
+        console.log(
+          `Telegram initialization failed, retrying in ${this.retryDelay}ms...`
+        );
+        await new Promise((resolve) => setTimeout(resolve, this.retryDelay));
```
stevenelleman, line 81: image

I intentionally did not set the env var to test this.  Will update the log so that it's explicit that we exited without initializing the bot.

stevenelleman: Updated:

image

apps/backend/src/routes/notification/index.ts
```diff
@@ -0,0 +1,49 @@
...
+      // Fetch user by auth token
+      const user = await controller.GetUserByAuthToken(authToken.toString());
+      if (!user) {
+        return res.status(401).json({ error: "Invalid auth token" });
+      }
+
+      const botUsername =
+        process.env.TELEGRAM_BOT_USERNAME ?? "CurtisCursiveBot";
+
+      const botLink = `https://t.me/${botUsername}?start=${user.id}`;
```
stevenelleman, line 39: Got it, so this is where the user provides the right ID for lookup, nice.

apps/backend/src/lib/controller/notification/telegram/client.ts
stevenelleman: Addressed: https://github.com/cursive-team/connections/pull/154/commits/7ce98da4ba460144b48f1f3580d3490aed651771

apps/backend/src/lib/controller/notification/telegram/client.ts
stevenelleman: Addressed: https://github.com/cursive-team/connections/pull/154/commits/7ce98da4ba460144b48f1f3580d3490aed651771

stevenelleman (approved):

apps/backend/src/lib/controller/notification/telegram/client.ts
stevenelleman: Previously after the max retries it would silently exit the method and enter the `.then` case in the api handler, which prints `Notification service initialized`, despite being not being properly set up.

PR #159, opened 2024-10-31
Halloween UI Fixes 

stevenelleman:


stevenelleman:
image

I made a profile change that initially seemed like a fix, but then it messed up the profile color. 

image

image

image

image

image

image

apps/frontend/src/components/ui/ProfileImage.tsx
```diff
@@ -8,6 +8,8 @@ const goudyBookletter = Goudy_Bookletter_1911({
 });
 
 export const getProfileBackgroundColor = (user: UserData) => {
+  const isHalloween: boolean = false;
```
stevenelleman, line 11: Didn't wind up doing this, because there was some kind of annoying passthrough on black, but not other colors. In theory I think this solution would work, which is why I kept it up -- we can quickly add new colors if need be.

RiverRuby (approved):

AndrewCLu (approved):

PR #163, opened 2024-11-02
Allow Chips to Be Added to Account 

stevenelleman:
Registered chips in local storage: 
Screenshot 2024-11-02 at 6 29 00 PM

Registered chips link to target account: 
https://github.com/user-attachments/assets/bb75fbc0-8356-40e4-b423-44e3f0b7b65b

### DIFF
--- a/apps/frontend/src/pages/tap.tsx
+++ b/apps/frontend/src/pages/tap.tsx
@@ -1,9 +1,16 @@
 import { useEffect } from "react";
 import { useRouter } from "next/router";
-import { TapParams, ChipTapResponse, errorToString, ChipIssuer } from "@types";
+import {
+  TapParams,
+  ChipTapResponse,
+  errorToString,
+  ChipIssuer,
+  JsonSchema
+} from "@types";
 import { toast } from "sonner";
 import { storage } from "@/lib/storage";
 import {
+  registerChip,
   tapChip,
   updateTapLeaderboardEntry,
   updateWorkoutLeaderboardEntry,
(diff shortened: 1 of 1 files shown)


apps/frontend/src/pages/tap.tsx
```diff
@@ -139,17 +146,33 @@ const TapPage: React.FC = () => {
...
-            // TODO: Allow users to register chips after account creation
-            logClientEvent("tap-chip-not-registered-logged-in", {});
-            toast.error(
-              "This chip is not registered yet. Please tell your friend to register it first!"
-            );
+          } else if (response.isLocationChip !== true && session && user?.userData) {
+            // User logged in and unregistered chip is not locationChip, allow user to bind new chip to profile
+            logClientEvent("tap-chip-logged-in-bind-new-chip", {});
+
+            const userData = JsonSchema.parse(user.userData);
```
AndrewCLu, line 156: this should be updated to use a function that selects the appropriate data for users, see comment below

AndrewCLu:
https://github.com/cursive-team/connections/blob/0713e4d5330acf0c0bba56cfb094717bb1c34a77/apps/frontend/src/lib/storage/localStorage/user/connection/message/tapBack.ts#L13

Just noticed: this needs to be updated to include new socials

AndrewCLu:
Can you hardcode this line to Lanna? https://github.com/cursive-team/connections/blob/0713e4d5330acf0c0bba56cfb094717bb1c34a77/apps/frontend/src/pages/community/index.tsx#L80

otherwise there is a dependence on the ordering of chips, which is confusing

stevenelleman:
> Can you hardcode this line to Lanna?
> 
> https://github.com/cursive-team/connections/blob/0713e4d5330acf0c0bba56cfb094717bb1c34a77/apps/frontend/src/pages/community/index.tsx#L80
> 
> otherwise there is a dependence on the ordering of chips, which is confusing

addressed [e20b7ea](https://github.com/cursive-team/connections/pull/163/commits/e20b7ea6ced0ffd8f553adc9a5cd617f259f1724)

apps/frontend/src/lib/storage/types/user/userData/index.ts
```diff
@@ -46,3 +46,23 @@ export {
...
+    encryptionPublicKey: userData.encryptionPublicKey,
+    psiPublicKeyLink: userData.psiPublicKeyLink,
+    twitter: userData.twitter,
+    telegram: userData.telegram,
+    signal: userData.signal,
+    instagram: userData.instagram,
+    farcaster: userData.farcaster,
+    pronouns: userData.pronouns,
+  };
+};
```
stevenelleman, line 68: Not sure this is the right place for this to live -- thoughts?

AndrewCLu: this can be under lib, maybe /lib/user

AndrewCLu (approved):
approved pending move of helper function to lib + testing

packages/types/src/util/index.ts
```diff
@@ -5,6 +5,7 @@ export const LiteralSchema = z.union([
   z.number(),
   z.boolean(),
   z.null(),
+  z.undefined(),
```
stevenelleman, line 8: @AndrewCLu this was added because the JsonSchema was panicking, currently we convert null to undefined in UserData but JsonSchema currently doesn't allow undefined. Do you know if this will have any weird consequences?

stevenelleman: JSON doesn't allow undefined so I found a better solution -- ie in this case convert undefined back to null.

PR #165, opened 2024-11-03
Add Password Strength Checker to Registration

stevenelleman:


stevenelleman:
Screenshot 2024-11-03 at 9 06 16 PM

Screenshot 2024-11-03 at 9 06 03 PM

stevenelleman:
Added more explicit first warning and cleaned up a bit: 
image

apps/frontend/package.json
```diff
@@ -48,6 +48,7 @@
     "eslint-config-next": "14.2.13",
     "postcss": "^8",
     "tailwindcss": "^3.4.1",
-    "typescript": "^5"
+    "typescript": "^5",
+    "zxcvbn": "^4.4.2"
```
AndrewCLu, line 52: does it make sense to use typescript version of this instead? https://www.npmjs.com/package/@zxcvbn-ts/core

apps/frontend/src/pages/register.tsx
```diff
@@ -166,6 +169,22 @@ const Register: React.FC = () => {
   };
 
   const handleRegisterWithPassword = async (password: string) => {
+
+    // Check password strength
+    const passwordCheck = zxcvbn(password);
+    if (passwordCheck && passwordCheck.score < 3) {
```
AndrewCLu, line 175: why 3? seems arbitrary, can you add a comment and link to some docs explaining this?

apps/frontend/src/pages/register.tsx
```diff
@@ -166,6 +169,22 @@ const Register: React.FC = () => {
...
+      let errorMsg: string = "Weak password, submit new one.";
+      if (passwordCheck.feedback?.warning) {
+        errorMsg += ` ${passwordCheck.feedback.warning}.`;
+      }
+      if (passwordCheck.feedback.suggestions.length > 0) {
```
AndrewCLu, line 180: i feel like feedback suggestions for a password, esp displayed as a toast, doesn't send the right ux message. this is more of a master password than an auth password, so we should just have the user retry

AndrewCLu:
> Screenshot 2024-11-03 at 9 06 16 PM Screenshot 2024-11-03 at 9 06 03 PM

This message is kinda jank. Can we instead just have a generic toast that says something like "use less common words, add numbers and symbols, etc."

apps/frontend/package.json
stevenelleman: Thanks for finding, looked and didn't find it originally.

stevenelleman:
> This message is kinda jank. Can we instead just have a generic toast that says something like "use less common words, add numbers and symbols, etc."

I don't think you pulled the most recent change. I fixed the styling: https://github.com/cursive-team/connections/pull/165#issuecomment-2454202250

Updated to remove suggestion, so it's not that relevant anyways: 
image

AndrewCLu (approved):
Update toast text to "Weak password, try adding numbers, symbols, and using less common words."

PR #173, opened 2024-11-05
Add mental health data and ability to Go Deeper

linamnt:
Allow users to "Go Deeper" and connect with users based on shared mental health and neurodivergent journeys. This will only reveal the intersection if both users choose to "Go Deeper" in the connection page.

Note: Not currently meant to actually be accepted, but more to get feedback on the POC
TODO: The "only reveal the intersection if both users...." is not exactly true right now because the private set intersection checks are lumped together with tensions/contacts. Ideally we split these out or parameterize it.


linamnt:
How it looks in app

Screenshot 2024-11-06 at 4 48 50 PM
Screenshot 2024-11-06 at 4 48 58 PM
Screenshot 2024-11-06 at 4 49 14 PM

RiverRuby:
Closing for now since the PR is quite outdated and adding this sort of data is not a roadmap priority. Thank you for your work @linamnt we will definitely revisit this in the future!

PR #175, opened 2024-11-05
Websocket Messaging

stevenelleman:
Quick demo: 
https://github.com/user-attachments/assets/cff5bea6-03ad-4034-8802-66cf473ed478

### DIFF
--- a/apps/backend/src/ws.ts
+++ b/apps/backend/src/ws.ts
@@ -67,6 +67,13 @@ wsServer.on('connection', (socket: WebSocket) => {
           }
 
           return;
+        case WebSocketRequestTypes.CLOSE:
+          if (!sender) {
+            return handleError(socket, sender, "Missing target.")
+          }
+          delete clientsSockets[sender];
+          socket.close();
+          return;
         default:
       };
     } catch (error) {
--- a/apps/backend/src/ws.ts
+++ b/apps/backend/src/ws.ts
@@ -1,0 +1,18 @@
+import { WebSocket } from 'ws';
+import { wsServer } from "@/index";
+
+wsServer.on('connection', (socket: WebSocket) => {
+
+  socket.on('open', () => {
+    console.log("Open ws client connection");
+  });
+
+  socket.on('message', async () => {
+    console.log("Receive message");
+  });
+
+  socket.on('close', () => {
+    console.log(`Close ws client connection`);
+  });
+
+});
--- a/apps/backend/src/ws.ts
+++ b/apps/backend/src/ws.ts
@@ -24,7 +24,7 @@ const handleError = (socket: WebSocket, target: string, error: string): void => 
     targetSigPubKey: target,
     senderSigPubKey: "",
     payload
-  }
+  };
 
   socket.send(JSON.stringify(resp))
   return;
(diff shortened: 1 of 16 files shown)


apps/backend/src/ws.ts
```diff
@@ -0,0 +1,117 @@
...
+  MapRequestToResponse,
+  errorToString
+} from "@types";
+import { Controller } from "@/lib/controller";
+
+import { wsServer } from "@/index";
+
+const controller = new Controller();
+
+const clientsSockets: Record<string, WebSocket> = {};
```
AndrewCLu, line 18: What are the implications of this being global, it just resets every server restart? Is there worry of this ever becoming too large?

apps/backend/src/ws.ts
```diff
@@ -0,0 +1,117 @@
...
+        case WebSocketRequestTypes.CONNECT:
+          if (!sender) {
+            return handleError(socket, sender, "Missing sender.");
+          }
+
+          // Set record for future lookup
+          clientsSockets[sender] = socket;
+
+          // Return CONNECT_SUCCESS response
+          const connectResponse: WebSocketResponse = MapRequestToResponse(req);
```
AndrewCLu, line 75: what is the purpose of the CONNECT_SUCCESS response?

apps/backend/src/ws.ts
```diff
@@ -0,0 +1,117 @@
...
+
+          if (clientsSockets[sender]) {
+            clientsSockets[sender].send(JSON.stringify(connectResponse));
+          } else {
+            throw new Error("Sender client socket missing.");
+          }
+
+          return;
+        case WebSocketRequestTypes.MSG:
```
AndrewCLu, line 84: i think its more descriptive to spell out MESSAGE, esp if we are using full words for the other types

apps/frontend/src/lib/user/index.ts
```diff
@@ -0,0 +1,11 @@
+import { User } from "@/lib/storage/types";
+
+// NOTE: other methods that depend on user object but not localstorage should be moved here
+export const getConnectionSigPubKey = (user: User, connectionUsername: string): string => {
```
AndrewCLu, line 4: this is expected to throw if connection is not found?

apps/frontend/src/lib/ws/index.ts
```diff
@@ -0,0 +1,141 @@
...
+  WebSocketErrorPayloadSchema,
+  CreateMessageDataSchema,
+  CreateMessageData,
+  MessageData,
+  MapCreateMessageDataToMessageData,
+} from "@types";
+
+// Global variables for connection status
+let expectConnectResp: boolean = false;
+let successfulConnection: boolean = false;
```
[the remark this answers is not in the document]
AndrewCLu: how are these global variables affected by connections resetting/dropping, etc?

apps/frontend/src/lib/ws/index.ts
```diff
@@ -0,0 +1,141 @@
...
+    const resp: WebSocketResponse = WebSocketResponseSchema.parse(JSON.parse(ev.data.toString()));
+
+    switch (resp.type) {
+      case WebSocketResponseTypes.SUCCESS_CONNECT:
+        // Enable other message types now
+        expectConnectResp = false;
+        successfulConnection = true;
+        return;
+      case WebSocketResponseTypes.MSG:
+        const createMessage: CreateMessageData = CreateMessageDataSchema.parse(resp.payload);
```
AndrewCLu, line 116: websocket messages should not be tied to the Message schema, they should be separate

stevenelleman:
Andrew and I just talked and this served a different intent than what Andrew was going for. Andrew wants a light "push to pull" notification system in which the client knows when to pull. Refactoring.

apps/frontend/src/lib/ws/index.ts
stevenelleman: Yup, this will be removed, it was involved because I thought WS was being used for low latency transfer, as opposed to notifying the client to pull.

PR #176, opened 2024-11-05
Refactor Data Imports and Extend Github Imports

stevenelleman:
In preparation for extending our data imports I refactored the code to be cleaner. I also added annual github contributions as a test case. Currently annual contributions are off.


apps/frontend/src/lib/imports/imports.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    for (const option of details.data_options) {
+      try {
+        // TODO: Need to specify data imports per community
+
+        // Unlike access token fetching, all data importing will be from client
+        await fetchAndSaveImportedData(
+          session.authTokenValue,
+          user,
+          ChipIssuer.EDGE_CITY_LANNA,
```
AndrewCLu, line 83: Can this be not hardcoded to Lanna?

apps/frontend/src/lib/imports/imports.ts
```diff
@@ -0,0 +1,96 @@
...
+import { fetchImportedData } from "./fetch";
+import { saveImportedData } from "./save";
+import { OAUTH_APP_DETAILS } from "@/config";
+import { storage } from "@/lib/storage";
+import { getOAuthAccessToken } from "@/lib/oauth";
+
+export async function fetchAndSaveImportedData(
+  authToken: string,
+  user: User,
+  chipIssuer: ChipIssuer,
```
AndrewCLu, line 18: For now maybe we can just remove chipIssuer and fetch this globally. Chip issuers can answer the question of "who does your data get shared with" post-Devcon, but for now we can just assume that data imports are global

AndrewCLu (approved):

apps/frontend/src/lib/imports/imports.ts
AndrewCLu: oh i see, this needs chip issuer to updated leaderboard, hmm

apps/frontend/src/lib/imports/imports.ts
AndrewCLu: nvm, i see saving is for updating leaderboard, hmm.

stevenelleman: In the next PR I generalize it so it can be supplied all the communities that need leaderboard entries.

apps/frontend/src/lib/imports/imports.ts
stevenelleman: Handle this here: https://github.com/cursive-team/connections/pull/176/files#r1832110685

apps/frontend/src/lib/imports/update.ts
```diff
@@ -0,0 +1,48 @@
...
+        },
+      };
+    case ImportDataType.GITHUB_CONTRIBUTIONS_LAST_YEAR:
+      const value3: number = data;
+      return {
+        ...userData,
+        github: {
+          annualCommits: {
+            value: value3,
+            lastUpdated: new Date()
```
stevenelleman, line 41: This should have a corresponding backup, will add in the next PR. This import type is not currently fetched (it was not added to the import options global var), will double check.

PR #177, opened 2024-11-06
Extend Telegram Bot Functionality

stevenelleman:
New features: 
- `/stop` command to pause notifications.
- `/start` will be used for both registration and for resuming notifications.
- `/feedback` will be used to link to a support channel. 
- `/help` list all commands 

Example: 
image

*In this example after ran `/stop` I tested that updating the user chip did not send a notification. After running `/start` again I also tested that the notification was correctly sent.

### DIFF
--- a/apps/backend/prisma/schema.prisma
+++ b/apps/backend/prisma/schema.prisma
@@ -23,6 +23,7 @@ model User {
   createdAt             DateTime    @default(now())
   notificationUserId    String?
   notificationUsername  String?
+  notificationsEnabled  Boolean?
   AuthToken             AuthToken[]
   Backup                Backup[]
 }
--- a/apps/backend/src/lib/controller/notification/telegram/client.ts
+++ b/apps/backend/src/lib/controller/notification/telegram/client.ts
@@ -1,6 +1,7 @@
 import { Bot } from "grammy";
 import { iNotificationClient } from "../interfaces";
 import { PrismaClient } from "@prisma/client";
+import { CommandContext, Context } from "grammy";
 
 export class TelegramNotificationClient implements iNotificationClient {
   private bot: Bot | undefined;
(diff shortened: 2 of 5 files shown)


apps/backend/src/lib/controller/notification/telegram/client.ts
```diff
@@ -18,49 +19,178 @@ export class TelegramNotificationClient implements iNotificationClient {
...
+      const startParameter = ctx.match;
+
+      if (startParameter) {
+        this.register(ctx);
+      } else {
+        this.resume(ctx);
+      }
+    });
+
+    this.bot.command("stop", async (ctx) => {
```
stevenelleman, line 165: I was waffling between `/stop` and `/pause` -- any preferences?

apps/backend/prisma/schema.prisma
```diff
@@ -23,6 +23,7 @@ model User {
   createdAt             DateTime    @default(now())
   notificationUserId    String?
   notificationUsername  String?
+  notificationsEnabled  Boolean
```
stevenelleman, line 26: Is this worth a migration or can / should I add it to the UserData?

stevenelleman: TODO: I need to add `@default(false)` to this.

AndrewCLu, line 26: This should be nullable, and make sure you delete the migration which added this as a nonnullable field otherwise we will delete all users lol

stevenelleman: oh jesus

stevenelleman: how often do we snapshot the db?

stevenelleman: Addressed: https://github.com/cursive-team/connections/pull/177/files#diff-8ce01239249acace801a09d07aaffa847a31687655eebfa2f3715a2f182bc4a7R26

stevenelleman: I'll now handle + backfill the null case.

AndrewCLu (approved):

PR #178, opened 2024-11-06
Private Tap Graph

stevenelleman:


packages/types/src/user/index.ts
```diff
@@ -17,6 +17,8 @@ export enum BackupEntryType {
   OAUTH = "OAUTH",
   LOCATION = "LOCATION",
   DELETE_OAUTH = "DELETE_OAUTH",
+  TOGGLE_SOCIAL_GRAPH = "TOGGLE_SOCIAL_GRAPH",
+  EDGE = "EDGE"
```
stevenelleman, line 21: The replace all didn't quite work, need to convert all `EDGE` sites to `TAP_GRAPH_EDGE`.

apps/frontend/src/lib/message/edge.ts
```diff
@@ -0,0 +1,48 @@
+import { z } from "zod";
+import {
+  MessageTypeSchema,
+  MessageType
+} from "@types";
+
+// TODO: need to update this
```
stevenelleman, line 7: Nit

packages/types/src/graph/index.ts
```diff
@@ -0,0 +1,34 @@
+import { z } from "zod";
+
+export const UpsertSocialGraphEdgeRequestQuerySchema = z.object({
```
stevenelleman, line 3: I should change this name

apps/backend/prisma/schema.prisma
```diff
@@ -68,6 +69,7 @@ model Message {
   senderEncryptedSignaturePublicKey  String
   senderEncryptedSignature           String
   createdAt                          DateTime @default(now())
+  type                               String? // Null type evaluates to TAP_BACK type
```
AndrewCLu, line 72: We should not add type to the `Message` table. See tap backs - the message type is included in the blob which is encrypted and sent to the other user. This ensure privacy at the message level

apps/backend/prisma/schema.prisma
```diff
@@ -135,3 +137,11 @@ model PairConnection {
   connectionScore Decimal?
   createdAt       DateTime @default(now())
 }
+
+model TapGraphEdge {
+    id          String @id @default(uuid())
+    tapperId      String? // Hash(Hash(Private Signing Key)) : use double hash so that single hash can be used for revocation
```
AndrewCLu, line 143: Let's use more clear terminology here, so one letter typos do not create bugs in the future - perhaps replace tapperId with tapSenderId and tappedId with tapReceiverId? this might be confusing as well, but we should come up with something else here

apps/backend/prisma/schema.prisma
stevenelleman: I was confused by that -- how do you know the type of the message so that you can properly parse it?

apps/backend/prisma/schema.prisma
stevenelleman: Reasonable. tapSender and tapReceiver is definitely better. Will keep thinking.

apps/backend/prisma/schema.prisma
AndrewCLu: you would decrypt the message, then you parse out the type field, and then based on that you parse it accordingly. this is not yet implemented, should be added here: https://github.com/cursive-team/connections/blob/a6748732dc8053ef564ef0205ea3e17e04985f7a/apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts#L49

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
```diff
@@ -138,14 +97,136 @@ export const processNewMessages = async (
...
   );
 
+  return newBackupEntries;
+}
+
+async function handleEdges(session: Session, user: User, messages: EdgeMessage[]): Promise<CreateBackupData[]> {
+  let tappedHash: string = "";
+  if (user && user.tapGraphEnabled) {
+    // Double hash the signature private key, use as identifier, use single hash version as revocation code
+    tappedHash = sha256(sha256(user.signaturePrivateKey));
```
AndrewCLu, line 107: same here, tappedHash = sha256(sha256(user.signaturePrivateKey) || "DEVCON2024");

apps/frontend/src/lib/backup/index.ts
```diff
@@ -473,3 +508,61 @@ export const createOAuthDeletionBackup = ({
...
+};
+
+export interface CreateEdgeArgs {
+  email: string;
+  password: string;
+  edgeId: string;
+  timestamp: Date;
+}
+
+export const createEdgeBackup = ({
```
AndrewCLu, line 544: don't we want to backup the username of the person this edgeId is for?

stevenelleman: I guess I don't see a downside of that and I do some future upsides (being able to label the global graph with your local "neighborhood"). Sure.

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
stevenelleman: Ah I see, is this so a single key can map to multiple issuer graphs?

AndrewCLu: yeah, i think the edgeId being universal is fine, but anything we do for the devcon graph should be specific to devcon

apps/backend/prisma/schema.prisma
stevenelleman: Addressed.

apps/backend/prisma/schema.prisma
stevenelleman: Addressed.

apps/frontend/src/features/graph/AddConsentButton.tsx
```diff
@@ -0,0 +1,75 @@
+import { FC, useEffect, useState } from "react";
+import { toast } from "sonner";
+import { storage } from "@/lib/storage";
+import { logClientEvent } from "@/lib/frontend/metrics";
+import { AppButton } from "@/components/ui/Button";
+import { createToggleSocialGraphBackup } from "@/lib/backup";
+import { saveBackupAndUpdateStorage } from "@/lib/storage/localStorage/utils";
+
+interface AddNotificationButtonProps {
```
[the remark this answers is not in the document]
stevenelleman: Addressed

apps/frontend/src/pages/tap.tsx
```diff
@@ -114,6 +117,26 @@ const TapPage: React.FC = () => {
             // Save tap to local storage
             await storage.addUserTap(response);
 
+            // If tap graph feature enabled, upsert graph edge (upsert so that either user can create the row)
+            let tapperHash: string = "";
+            if (user && user.tapGraphEnabled) {
+              // Double hash the signature private key, use as identifier, use single hash version as revocation code
+              tapperHash = sha256(sha256(user.signaturePrivateKey));
```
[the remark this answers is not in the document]
stevenelleman: Addressed

apps/frontend/src/lib/backup/index.ts
stevenelleman: Addressed (added username to local storage / backup)

apps/frontend/src/lib/backup/index.ts
```diff
@@ -473,3 +501,66 @@ export const createOAuthDeletionBackup = ({
...
+  return {
+    authenticationTag,
+    iv,
+    encryptedData,
+    backupEntryType: BackupEntryType.TOGGLE_SOCIAL_GRAPH,
+    clientCreatedAt: new Date(),
+  };
+};
+
+export interface CreateEdgeArgs {
```
AndrewCLu, line 529: i think there should also be a notion here of "isTapSender"? seems like since edges are directed its important to know which side you are

AndrewCLu: although i guess the server could provide you this information?

AndrewCLu: actually this might be a better reason to store directionality in the backup. so technically this feature could be composable across different servers, i.e. you can upload your entire tap graph to a different server in the future

apps/frontend/src/features/graph/AddConsentButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+      setButtonText(user.tapGraphEnabled ? 'Disable Graph' : 'Enable Graph');
+      setUpdatedSetting(false);
+    }
+    checkSetting();
+  }, [updatedSetting]);
+
+  const handleGraphConsentSetup = async () => {
+    logClientEvent("toggle-tap-graph", {});
+
+    const { user, session } = await storage.getUserAndSession();
```
AndrewCLu, line 39: this entire function should probably be in storage, i.e. storage.toggleSocialGraphConsent

apps/frontend/src/lib/message/edge.ts
```diff
@@ -0,0 +1,49 @@
...
+export const EdgeMessageSchema = z.object({
+  type: MessageTypeSchema,
+  edge: EdgeMessageContentSchema,
+});
+
+export type EdgeMessage = z.infer<typeof EdgeMessageSchema>;
+
+export interface GenerateEdgeMessageArgs {
+  edgeId: string,
+  senderUsername: string,
```
AndrewCLu, line 18: what is the role of `senderUsername` in this message?

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
```diff
@@ -138,14 +98,137 @@ export const processNewMessages = async (
...
   );
 
+  return newBackupEntries;
+}
+
+async function handleEdges(session: Session, user: User, messages: EdgeMessage[]): Promise<CreateBackupData[]> {
+  let tapReceiverHash: string | null = null;
+  if (user && user.tapGraphEnabled) {
+    // Double hash the signature private key, use as identifier, use single hash version as revocation code
+    tapReceiverHash = sha256(sha256(user.signaturePrivateKey) || DEVCON);
```
AndrewCLu, line 108: this should be string concat, not OR

apps/frontend/src/pages/tap.tsx
```diff
@@ -114,6 +118,30 @@ const TapPage: React.FC = () => {
             // Save tap to local storage
             await storage.addUserTap(response);
 
+            // If tap graph feature enabled, upsert graph edge (upsert so that either user can create the row)
+            let tapSenderHash: string | null = null;
+
+            const sentHash: boolean = (user?.tapGraphEnabled === true);
+            if (sentHash) {
+              // Double hash the signature private key, use as identifier, use single hash version as revocation code
+              tapSenderHash = sha256(sha256(user.signaturePrivateKey) || DEVCON);
```
AndrewCLu, line 127: this should be string concat, not OR

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
```diff
@@ -138,14 +98,137 @@ export const processNewMessages = async (
...
+          connectionTapBacks[senderSignaturePublicKey] = [
+            ...connectionTapBacks[senderSignaturePublicKey],
+            tapBackMessage,
+          ];
+        } else {
+          connectionTapBacks[senderSignaturePublicKey] = [tapBackMessage];
+        }
+        continue;
+      } else if (serializedData.match(MessageType.TAP_GRAPH_EDGE.toString())) {
+        const edgeMessage = parseSerializedEdgeMessage(serializedData);
```
AndrewCLu, line 192: as a future note, here we should check that the `connectionUsername` in the edge message matches the `signaturePublicKey`. this is not an easy fix in the current setup since at this point the connection struct might not exist, but would be doable with a public registry

apps/frontend/src/pages/tap.tsx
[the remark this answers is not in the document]
stevenelleman: > as discussed in chat, let's have this be `tapperHash = sha256(sha256(user.signaturePrivateKey) || "DEVCON2024");`

^ in this context does || just mean concat?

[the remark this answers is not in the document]
AndrewCLu: yes concat, confusing sorry

apps/frontend/src/lib/backup/index.ts
stevenelleman: added

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
stevenelleman: signaturePublicKey in your connections object?

apps/frontend/src/lib/message/edge.ts
stevenelleman: The other user needs your name to set in their local edge info.

apps/backend/src/lib/controller/graph/types.ts
```diff
@@ -0,0 +1,9 @@
+
+
+export type EdgeData = {
```
RiverRuby, line 3: Any reason this isn't a proper schema?

apps/frontend/src/pages/tap.tsx
```diff
@@ -114,6 +118,36 @@ const TapPage: React.FC = () => {
...
+
+            const sentHash: boolean = (user?.tapGraphEnabled === true);
+            if (sentHash) {
+              // Double hash the signature private key, use as identifier, use single hash version as revocation code
+              tapSenderHash = sha256(sha256(user.signaturePrivateKey).concat(DEVCON));
+            }
+
+            // Upsert row, returns edge ID
+            try {
+              const resp: UpsertSocialGraphEdgeResponse = await upsertSocialGraphEdge(session.authTokenValue, null, tapSenderHash, null);
```
RiverRuby, line 132: why are we upserting with null tapSenderHashs?

apps/frontend/src/pages/profile/index.tsx
```diff
@@ -135,6 +136,9 @@ const ProfilePage: React.FC = () => {
                     <span className="text-[14px]">Add Chip</span>
                   </AppButton>
                 </div>
+                <div className="w-fit">
```
RiverRuby, line 139: Can this be near the bottom (near dark theme and logout) instead?

RiverRuby (approved):
LGTM after comments are addressed. From what I understand there's no backfilling?

apps/frontend/src/lib/backup/index.ts
```diff
@@ -155,6 +156,33 @@ export const processUserBackup = ({
           delete user.oauth[oauthApp.app];
         }
         break;
+      case BackupEntryType.TOGGLE_SOCIAL_GRAPH:
+        if (!user) {
+          throw new Error("TOGGLE_SOCIAL_GRAPH backup entry found before INITIAL");
```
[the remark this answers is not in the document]
stevenelleman: every case has a throw error?

apps/frontend/src/pages/tap.tsx
stevenelleman: (1) because each side can only submit their own hash. 
(2) each side can choose to consenting to sending their hash. If they don't consent, we still create the edge so they can still add them later.

apps/frontend/src/pages/profile/index.tsx
stevenelleman: Fixed

apps/backend/src/lib/controller/graph/types.ts
stevenelleman: fixed.

PR #185, opened 2024-11-08
Delete OAuth Expired Token

stevenelleman:


apps/frontend/src/lib/imports/imports.ts
```diff
@@ -175,7 +176,8 @@ export async function refreshData(): Promise<void> {
 
         if (accessToken.expires_at && now > expiresAt) {
           // TODO: swap to use refresh token
-          toast.error(`${app.toString()} token has expired, go through OAuth flow again`);
+          storage.deleteOAuthAccessToken(option.type);
+          toast.error(`${capitalized} token has expired, go through OAuth flow again.`, {duration: 5000});
```
stevenelleman, line 181: Right now an expired token leads to a loop where continually reattempt to reimport with the expired token. By deleting it from localstorage I believe it should go into the case on L164 (which returns a console error but not an error toast).

AndrewCLu:
does this include the fix where toasts should not appear when oauth fetches happen in the background? only makes sense to display success/error if the user manually performs an action

stevenelleman:
> does this include the fix where toasts should not appear when oauth fetches happen in the background? only makes sense to display success/error if the user manually performs an action

will remove all toasts from this flow.

stevenelleman:
> does this include the fix where toasts should not appear when oauth fetches happen in the background? only makes sense to display success/error if the user manually performs an action

I only keep in cases where the toast will occur once, in the case of expired and revoked token we now delete the token and show the toast to inform the user to go through the oauth flow again. In the future when we use the refresh token the user won't have to go through the oauth flow again.

AndrewCLu (approved):

PR #191, opened 2024-11-09
Socket Tapbacks 

stevenelleman:
Short Demo: 

https://github.com/user-attachments/assets/11d6fe97-6c81-4fd5-bc16-e69845ac9122


apps/backend/src/routes/socket/index.ts
```diff
@@ -0,0 +1,120 @@
...
+    senderSigPubKey: "",
+    payload
+  };
+
+  socket.send(JSON.stringify(resp))
+  return;
+}
+
+wsServer.on('connection', (socket: Socket) => {
+  socket.on('open', () => {
```
AndrewCLu, line 38: not urgent, but does socket.io have authenticated connections? so auth token doesnt need to be sent with every message?

apps/backend/src/routes/socket/index.ts
```diff
@@ -0,0 +1,120 @@
...
+      }
+
+      sender = req.senderSigPubKey;
+      recipient = req.recipientSigPubKey;
+
+      if (!req.authToken) {
+        return handleError(socket, sender, "Missing auth token.");
+      }
+
+      // Fetch user by auth token, ensure the request is from an authenticated user
```
AndrewCLu, line 63: Not urgent, but if we're doing this verification then the user does not need to send their `senderSigPubKey` with the request

apps/backend/src/routes/socket/index.ts
```diff
@@ -0,0 +1,120 @@
...
+      }
+
+      switch (req.type) {
+        case SocketRequestType.CONNECT:
+          if (!sender) {
+            return handleError(socket, sender, "Missing sender.");
+          }
+
+          // Set record for future lookup
+          clientsSockets[sender] = socket;
```
AndrewCLu, line 78: one thing to test here is multiple tabs - how does opening two tabs affect the socket connection?

apps/backend/src/routes/socket/index.ts
```diff
@@ -0,0 +1,120 @@
...
+          // Return CONNECT_SUCCESS response
+          const connectResponse: SocketResponse = MapRequestToResponse(req);
+
+          if (clientsSockets[sender]) {
+            clientsSockets[sender].send(JSON.stringify(connectResponse));
+          } else {
+            throw new Error("Sender client socket missing.");
+          }
+          return;
+        case SocketRequestType.TAP_BACK:
```
AndrewCLu, line 89: Again not urgent, but in the future having the sender send a tap back notif via websockets means the server can see this activity. a more private flow is that the server just notifies a client itself when it sees it has a new message (where sender privacy is provided via ephemeral ecdh)

apps/frontend/src/lib/socket/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+const SocketContext = createContext({} as SocketContextProps);
+
+export const SocketProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
+  const [socket, setSocket] = useState<Socket | null>(null);
+  const [connected, setConnected] = useState<boolean>(false);
+
+  useEffect(() => {
+    // HERE: fix this:
```
AndrewCLu, line 35: ?

apps/frontend/src/lib/socket/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  return (
+    <SocketContext.Provider value={{ socket, connected }}>
+      {children}
+    </SocketContext.Provider>
+  );
+};
+
+// Custom hook to use socket context
+export const useSocket = (): SocketContextProps => {
```
AndrewCLu, line 116: sick

apps/frontend/src/pages/_app.tsx
```diff
@@ -34,6 +36,7 @@ export default function MyApp({ Component, pageProps }: AppProps) {
   const router = useRouter();
   const [isPreMigrationSessionChecked, setIsPreMigrationSessionChecked] =
     useState(false);
+  const [runRefreshMessageLoop, setRunRefreshMessageLoop] = useState(true);
```
AndrewCLu, line 39: where is this used?

AndrewCLu (approved):
looks great! context + hook is super clean. would just test the flow with multiple tabs open to make sure everything works (since tapping on mobile browser will open multiple tabs)

apps/frontend/src/lib/socket/helper.ts
```diff
@@ -0,0 +1,127 @@
+import { Socket } from 'socket.io-client';
```
stevenelleman, line 1: @AndrewCLu accidentally didn't push this file in my last push.

apps/frontend/src/pages/_app.tsx
stevenelleman: Isn't used, removed.

apps/backend/src/routes/socket/index.ts
stevenelleman: addressed [de28bf9](https://github.com/cursive-team/connections/pull/191/commits/de28bf9aee23d2bbaba9ddcc10b8f97649e14d06)

Commit message: 
```
a commit of beauty: in providing the auth token, the user can be looked up and the pubkey-to-socket lookup entry can be set up. This allows the CONNECT-CONNECT_SUCCESS step to be removed as an established connection already implies that lookup has been set up. This also allows the sender to be removed from all future requests, as the information has been attached to the user's associated socket. Glorious.
```

apps/backend/src/routes/socket/index.ts
stevenelleman: https://github.com/cursive-team/connections/pull/191#discussion_r1835542731

apps/backend/src/routes/socket/index.ts
stevenelleman: Tested -- I sent a tapback while there were 3 tabs that were all connected to the same user profile, only the most recently reloaded tab got the tapback and the other 2 tabs didn't respond to the change (no update, no error). This seems like acceptable behavior to me.

apps/backend/src/routes/socket/index.ts
stevenelleman: Unfortunately I don't understand the alternative you're suggesting. Is the idea that the server would poll for new messages, and when it gets a new message it would see if there's an active client socket, else skip?

apps/backend/src/routes/socket/index.ts
AndrewCLu: yup thats fine

apps/backend/src/routes/socket/index.ts
AndrewCLu: Nah, when the server receives new messages, it just sends a ws ping to all the recipients. but this is not necessary for now

AndrewCLu (approved):
LGTM

apps/backend/src/routes/socket/index.ts
stevenelleman: https://github.com/cursive-team/connections/pull/191/files#diff-e1f99bd817c2c6047065d9e66074644a68ab0e614021785848519a46c26cf730R99 -- isn't this only sending it to the recipient socket?

PR #192, opened 2024-11-09
remove email and code from registration flow

AndrewCLu:
- removes email and signing code from registration
- disables lanna registration
- updates login to use username

NOTE: now we use email in database and localStorage to actually be a backup salt. For users before this change, the backup salt will actually be their verified email, for users after the backup salt will be their username. We will update naming and possibly ensure better compatibility in a future change, but for now this enables us to easily make the email optional as it was previously required as a backup salt.


stevenelleman:
image

stevenelleman:
Flow for new account worked.

apps/frontend/src/lib/storage/types/user/index.ts
```diff
@@ -3,15 +3,11 @@ import { UserDataSchema } from "./userData";
...
-  AccessTokenSchema,
-  EmailSchema,
-  nullToUndefined
-} from "@types";
+import { AccessTokenSchema, nullToUndefined } from "@types";
 import { LocationSchema } from "./location";
 
 export const UserSchema = z.object({
-  email: EmailSchema,
+  email: z.string(), // Email here is actually used as a backup salt for backwards compatibility reasons.
```
stevenelleman, line 10: ty for adding context.

apps/frontend/src/pages/login.tsx
```diff
@@ -110,7 +81,7 @@ const LoginPage: React.FC = () => {
 
     try {
       setLoading(true);
-      await processLoginResponse(loginResponse, email, password);
+      await processLoginResponse(loginResponse, loginResponse.email, password);
```
stevenelleman, line 84: When I wake up from a nap I'm going to check if loginResponse.email exists. It felt like email has been removed from most locations and replaced with username.

apps/frontend/src/features/login/EnterUsername.tsx
```diff
@@ -8,39 +8,45 @@ import { AppCopy } from "@/components/ui/AppCopy";
...
+  const [username, setUsername] = useState("");
   const [loading, setLoading] = useState<boolean>(false);
 
   const handleSubmit = async (e: React.FormEvent) => {
     e.preventDefault();
+    try {
+      UsernameSchema.parse(username);
+    } catch (error) {
+      console.error(error);
+      toast.error("Please enter a valid username");
```
stevenelleman, line 25: I liked the other error message slightly better because it's more explanatory (`"Username must be alphanumeric and between 3-20 characters"`).

apps/frontend/src/features/register/devcon/Register.tsx
```diff
@@ -194,8 +135,7 @@ const RegisterDevcon: React.FC<RegisterDevconProps> = ({ savedTap }) => {
     try {
       // Register user
       await registerUser({
-        signinToken: code,
-        email,
+        email: username,
```
stevenelleman, line 138: I bet this is where Response.email is coming from (so ignore my earlier comment).

stevenelleman:
image

Testing backwards compatibility with Lanna user created before this PR.

apps/frontend/src/features/login/EnterUsername.tsx
stevenelleman: Or is this more vague because it can also take emails?

stevenelleman (approved):
Was able to log in with an old lanna user and a new user. LGTM.

PR #193, opened 2024-11-10
Refresh Token in the Background when Access Token Expires

stevenelleman:


apps/frontend/src/lib/imports/imports.ts
```diff
@@ -186,12 +186,17 @@ export async function refreshData(): Promise<void> {
...
+          // If token expired, use refresh token, else warn user and delete token
+          try {
+            accessToken = await refreshAndSaveToken(app, details, accessToken);
+          } catch (error) {
+            const tokenExists = await storage.getOAuthAccessToken(app);
+            if (tokenExists) {
+              storage.deleteOAuthAccessToken(app);
+              // As this toast will only be shown once, keep it
+              console.error("Token refresh failed:", errorToString(error));
+              toast.error(`${capitalized} token has expired, reauth app to refresh token.`, {duration: 5000});
```
AndrewCLu, line 198: Does this happen in the background or on manual action? If background then we should not display this toast, per previous discussion

AndrewCLu (approved):
this looks fine for this change, but we should have a future refactor that checks to see if refreshes are being triggered from a user action or background process, and if background process then it should fail gracefully/never toast

PR #198, opened 2024-11-11
Implement new connections page + flower garden

kalidiagne:


stevenelleman:
Otherwise looks fine to me and works locally.

apps/frontend/src/pages/people/[username].tsx
```diff
@@ -475,6 +565,15 @@ const UserProfilePage: React.FC = () => {
     return;
   }
 
+  let size = "sprout";
+  if (user?.userData?.connectionPSISize && user?.userData?.connectionPSISize[connection.user.username]) {
+    size = flowerSize(user?.userData?.connectionPSISize[connection.user.username]);
+  }
+
+  const flowerStage = size;
+  const flowerIndex = "2";
```
Comment (RiverRuby) [inline apps/frontend/src/pages/people/[username].tsx:574]: Needs be mod 10

apps/frontend/tailwind.config.ts
```diff
@@ -45,6 +45,8 @@ const config: Config = {
         quaternary: "var(--color-quaternary)",
         cinnamon: "var(--color-cinnamon)",
 
+        pink: "#ff9df8",
```
RiverRuby, line 48: Unideal but we can fix later

RiverRuby: geneerally using variabes is really good when we want to do things like dark mode, i'd try to find the right variable here

RiverRuby (approved):
LGTM minus nits

PR #199, opened 2024-11-11
Simplify Server Sockets Further and Add Logging

stevenelleman:


apps/backend/src/routes/socket/index.ts
```diff
@@ -26,7 +24,7 @@ export const wsServer = new Server(server);
 
 const controller = new Controller();
 
-const clientsSockets: Record<string, Socket> = {};
+const clientsSockets: Record<string, string> = {};
```
stevenelleman, line 27: Switch to storing pub key to socket id. Perhaps socket.io changes sockets in the background while persisting their IDs? That would really mess up my original implementation, which assumes I can use the same socket object.

RiverRuby (approved):
LGTM, seems like it won't mess up other stuff

PR #205, opened 2024-11-11
Full Accountless Flow

stevenelleman:


apps/frontend/src/pages/people/index.tsx
```diff
@@ -17,17 +28,17 @@ const PeoplePage: React.FC = () => {
   useEffect(() => {
     const fetchConnections = async () => {
       const user = await storage.getUser();
-      if (!user) {
+      const unregisteredUser = await getUnregisteredUser();
+      if (!user && !unregisteredUser) {
         return;
```
RiverRuby, line 33: Should this router push somewhere else?

apps/frontend/src/pages/tap.tsx
```diff
@@ -63,15 +63,6 @@ const TapPage: React.FC = () => {
               return;
             }
 
-            // If user is not logged in, direct them to sign in
```
RiverRuby, line 66: If we're not doing location taps in accountless then why did we remove this?

apps/frontend/src/pages/tap.tsx
```diff
@@ -114,8 +96,11 @@ const TapPage: React.FC = () => {
             // Save tap to local storage
             await storage.addUserTap(response);
```
RiverRuby, line 97: shouldn't this not happen if user is not registered?

apps/frontend/src/pages/tap.tsx
```diff
@@ -114,8 +96,11 @@ const TapPage: React.FC = () => {
...
-            // Update leaderboard entry
-            await updateTapLeaderboardEntry(response.chipIssuer);
+            if (user) {
+              // If no user, skip updating tap leaderboard.
+              // For the team being assume unregistered users won't be recorded
+              await updateTapLeaderboardEntry(response.chipIssuer);
+            }
 
             // Save tap to populate modal upon redirect
             await storage.saveTapInfo({ tapParams, tapResponse: response });
```
RiverRuby, line 106: same with this?

apps/frontend/src/pages/people/index.tsx
RiverRuby, line 33: shouldn't we boot them off this page in this case?

apps/frontend/src/pages/people/[username].tsx
```diff
@@ -206,34 +206,57 @@ const UserProfilePage: React.FC = () => {
...
+            savedTapInfo.tapResponse.userTap?.ownerUsername === username
+          ) {
+            logClientEvent("user-profile-tap-chip-modal-shown", {});
+            setTapInfo(savedTapInfo);
+            setShowCommentModal(true);
+          }
+        } else if (unregisteredUser) {
+          // In unregistered user case, cannot do PSIs which require user object
+          setConnection(unregisteredUser.connections[username]);
+          const savedTapInfo = await storage.loadSavedTapInfo();
```
Comment (RiverRuby) [inline apps/frontend/src/pages/people/[username].tsx:248]: I see, the earlier saveTpa it's used to transfer over info to this page, makes sense

apps/frontend/src/pages/tap.tsx
RiverRuby: i see this is used to get info to the people page

RiverRuby (commented):
I'm feeling a bit confused on the tap.tsx flow, seems a bit unprotected especially re: locationTaps. Which will exist in the museum so we should make sure that doesn't get user into a weird state

apps/frontend/src/lib/auth/register.ts
```diff
@@ -141,3 +166,77 @@ export async function registerUser({
...
+
+  const user: UnregisteredUser = {
+    signaturePrivateKey,
+    encryptionPrivateKey,
+    lastMessageFetchedAt: new Date(),
+    userData: {
+      signaturePublicKey,
+      encryptionPublicKey,
+    },
+    connections: {},
```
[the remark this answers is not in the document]
stevenelleman: I'll make an empty record for locations.

apps/frontend/src/pages/tap.tsx
stevenelleman: Adding back in.

apps/frontend/src/lib/auth/register.ts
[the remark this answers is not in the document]
stevenelleman: Good catch.

apps/frontend/src/pages/people/index.tsx
stevenelleman: Good catch.

apps/frontend/src/lib/storage/types/user/index.ts
```diff
@@ -22,6 +26,21 @@ export const UserSchema = z.object({
 
 export type User = z.infer<typeof UserSchema>;
 
+export const UnregisteredUserSchema = z.object({
+  signaturePrivateKey: z.string(),
+  encryptionPrivateKey: z.string(),
+  lastMessageFetchedAt: z.coerce.date(),
+  userData: UnregisteredUserDataSchema,
+  connections: z.record(z.string(), ConnectionSchema),
+  locations: nullToUndefined(z.record(z.string(), LocationSchema)),
```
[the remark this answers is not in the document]
stevenelleman: out of scope.

apps/frontend/src/lib/auth/register.ts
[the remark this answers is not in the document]
stevenelleman: added, addressed, ty.

apps/frontend/src/pages/tap.tsx
stevenelleman: added back, tested and it works, addressed

apps/frontend/src/pages/tap.tsx
stevenelleman: we need this for adding the connection to the unregistered user

apps/frontend/src/pages/people/index.tsx
stevenelleman: I now boot, addressed

RiverRuby (approved):
LGTM

PR #213, opened 2024-11-11
Only Execute Proofs If User Exists and Is Logged In 

stevenelleman:


apps/frontend/src/pages/login.tsx
```diff
@@ -82,6 +82,9 @@ const LoginPage: React.FC = () => {
     try {
       setLoading(true);
       await processLoginResponse(loginResponse, loginResponse.email, password);
+
+      // If you have a proper login, clean up localstorage
+      await storage.deleteUnregisteredUser();
```
stevenelleman, line 87: Tested -- it occurred to me when I tested my own account is production that after making an unregistered user and then logging into my account the unregistered account would still exists. This ensures that on a login an unregistered user would be deleted from localstorage.

RiverRuby (approved):
LGTM

PR #222, opened 2024-11-13
Add Edge Backfill on Graph Feature Enablement 

stevenelleman:


apps/frontend/src/lib/storage/localStorage/graph.ts
```diff
@@ -9,6 +13,31 @@ export const toggleGraphConsent = async (): Promise<boolean> => {
     session.authTokenValue &&
     session.authTokenExpiresAt > new Date()
   ) {
+
+    // The value is updated in the backup method, so it should be currently false and will soon be flipped to true
+    if (!user.tapGraphEnabled && user.edges) {
```
micahscopes, line 18: I don't understand why this is `!user.tapGraphEnabled`

micahscopes: Ahh, I see, this function is for when people enable it for the first time, not for toggling in both directions (yet)

RiverRuby (approved):

PR #226, opened 2024-11-13
Save chipId to Connection Via Tap, Surface Only for Admins

stevenelleman:


apps/frontend/src/lib/storage/types/user/connection/index.ts
```diff
@@ -12,6 +12,7 @@ export const ConnectionSchema = z.object({
   taps: z.array(TapDataSchema),
   psi: nullToUndefined(PSIDataSchema),
   sentMessages: nullToUndefined(z.array(SentMessageSchema)).default([]),
+  chipId: z.string().optional(),
```
stevenelleman, line 15: Is optional the proper type for when a value may start as undefined?

apps/frontend/src/lib/storage/types/user/tap.ts
```diff
@@ -0,0 +1,13 @@
+import { z } from "zod";
+import { ChipIssuerSchema, LocationTapSchema, UserTapSchema } from "@types";
+
+export const ChipTapSchema = z.object({
+  chipIssuer: ChipIssuerSchema,
+  chipIsRegistered: z.boolean(),
+  isLocationChip: z.boolean().nullable(),
+  userTap: UserTapSchema.nullable(),
+  locationTap: LocationTapSchema.nullable(),
+  chipId: z.string(),
```
stevenelleman, line 10: Just adds the chipId to the ChipTapResponseSchema.

RiverRuby (approved):

PR #237, opened 2024-11-26
PSI With Automatic Refreshing

stevenelleman:


apps/frontend/src/pages/people/[username].tsx
```diff
@@ -747,12 +630,12 @@ const UserProfilePage: React.FC = () => {
                   ) : (
                     <div className="text-sm text-link-primary font-sans font-normal">
                       {verifiedIntersection.contacts.map((contact, index) => (
-                        <>
+                        <div key={index}>
```
Comment (stevenelleman) [inline apps/frontend/src/pages/people/[username].tsx:633]: Fixes warning unrelated to other changes.

apps/frontend/src/pages/people/[username].tsx
```diff
@@ -361,169 +376,37 @@ const UserProfilePage: React.FC = () => {
...
-            },
-          }),
-        }
-      );
-
-      if (!response.ok) {
-        const errorData = await response.json();
-        throw new Error(`HTTP error! status: ${errorData.error}`);
-      }
+      const newVerifiedIntersection = await refreshPSI(user, connection);
```
Comment (stevenelleman) [inline apps/frontend/src/pages/people/[username].tsx:385]: Deleted code moved into helper method. Helper method is used both by clicking Discover / Refresh button and by automatic PSI triggered by socket push notification.

apps/frontend/src/pages/people/[username].tsx
```diff
@@ -275,7 +290,7 @@ const UserProfilePage: React.FC = () => {
     };
 
     fetchConnectionAndTapInfo();
-  }, [username, router]);
+  }, [username, router, user]);
```
Comment (stevenelleman) [inline apps/frontend/src/pages/people/[username].tsx:293]: Added user dependency for when `user.userData.PullPSI` is updated by `refreshMessages`. When the user is updated it means refreshPSI was run, and we need to pick up the verified intersection so it can automatically populate the page.

apps/frontend/src/lib/storage/types/user/userData/index.ts
```diff
@@ -32,6 +38,7 @@ export const UserDataSchema = z.object({
   lannaHalloween: nullToUndefined(LannaHalloweenDataSchema),
   devcon: nullToUndefined(DevconSchema),
   connectionPSISize: nullToUndefined(ConnectionPSISizeSchema),
+  pullPSI: nullToUndefined(PullPSISchema),
```
stevenelleman, line 41: Adding to keep track of when a connection should automatically refresh their PSI results. I don't love the name, probably will change it tomorrow.

apps/frontend/src/lib/storage/localStorage/user/connection/message/psi.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const connection = user.connections[connectionUsername];
+  if (!connection) {
+    throw new Error(`Connection not found for username: ${connectionUsername}`);
+  }
+
+  const { serializedMessage, messageTimestamp } =
+    await generateSerializedPSIMessage({
+      // Send sender username to recipient
+      senderUsername,
```
stevenelleman, line 22: You need to know the name to know which connection to refresh.

apps/frontend/src/lib/psi/refresh.ts
```diff
@@ -0,0 +1,206 @@
...
+    if (user.userData?.devcon?.schedule) {
+      devconEventTitles = user.userData.devcon.schedule.map(event => event.title);
+      devconEvents = await hashCommit(
+        user.encryptionPrivateKey,
+        connection.user.encryptionPublicKey,
+        devconEventTitles
+      );
+    }
+
+    // HERE: any ability to order on frequency?
```
stevenelleman, line 56: Need to remove, this is from a prior PR.

apps/backend/src/index.ts
```diff
@@ -159,36 +159,49 @@ wsServer.on("connection", (socket: Socket) => {
...
+            socket.to(socketId).emit(SocketResponseType.PULL_PSI_MSG);
+          } else {
+            // If client is not online, use telegram instead
+            const targetUser: User | null = await controller.GetUserBySigPubKey(
+              recipient
+            );
+            if (targetUser) {
+              await controller.SendNotification(
+                targetUser.id,
+                `${socket.user.username} has refreshed your intersection! Go to their contact card and check out what common ground you share.`
```
stevenelleman, line 202: I should add an embedded link to the connection contact page.

stevenelleman:
I'll rebase this soon.

stevenelleman:
I need to add a user-controlled setting for automatic PSI.

apps/frontend/src/lib/psi/refresh.ts
```diff
@@ -0,0 +1,242 @@
...
+  Connection,
+  UserData,
+  Session
+} from "@/lib/storage/types";
+import { storage } from "@/lib/storage";
+import { sendMessages } from "@/lib/message";
+import { Socket } from "socket.io-client";
+import { updateUserData } from "@/lib/storage/localStorage/user/userData";
+
+export type Intersection = {
```
AndrewCLu, line 24: This should eventually be a storage-level type once we are saving PSI results

apps/frontend/src/lib/psi/refresh.ts
```diff
@@ -0,0 +1,242 @@
...
+        user.encryptionPrivateKey,
+        connection.user.encryptionPublicKey,
+        tensionData
+      );
+    }
+
+    let hotTakes: string[] = [];
+    if (user.userData.hotTakesRating?.revealAnswers) {
+      const hotTakeData = hotTakeLabels.map((hotTake, index) => {
+        switch (user.userData.hotTakesRating!.rating[index]) {
```
AndrewCLu, line 52: Little confused how this works - doesn't there need to be a unique string in the hash commit, such as the name of the hotTake, so that the hashes are actually unique?

apps/frontend/src/lib/storage/types/user/userData/index.ts
AndrewCLu: if this is related to connections refreshing psi results why is it at the userData struct level?

apps/frontend/src/lib/psi/refresh.ts
stevenelleman: Agreed, as I mentioned in TG want to add that in the plaintext PSI / friend PSIs where the PSI info is actually used.

stevenelleman:
> if this is related to connections refreshing psi results why is it at the userData struct level? 

Yeah good point, I'm adding a UserSettings object for the automatic PSI setting, it may make sense to expand the scope of the object a bit to include PullPSI. 

Also small nit, but I can't respond to specific comments in Github reviews. I have no idea why GH does this, seems silly to me. This is all to say that I have a slight preference towards individual comments instead of GH reviews.

apps/frontend/src/lib/psi/refresh.ts
stevenelleman: I'll look into this in the Github PSI PR.

apps/frontend/src/lib/storage/types/user/userData/index.ts
```diff
@@ -40,7 +50,7 @@ export const UserDataSchema = z.object({
   lannaHalloween: nullToUndefined(LannaHalloweenDataSchema),
   devcon: nullToUndefined(DevconSchema),
   connectionPSISize: nullToUndefined(ConnectionPSISizeSchema),
```
stevenelleman, line 52: I think this was the wrong place to put the PSI size -- I think it may make more sense to put it with the connection itself.

stevenelleman: While I like using connection more, I feel like there's enough going on in this PR. I have the commit, but I'll probably make it into a follow up PR.

apps/frontend/src/lib/storage/types/user/connection/index.ts
```diff
@@ -12,6 +12,7 @@ export const ConnectionSchema = z.object({
   taps: z.array(TapDataSchema),
   psi: nullToUndefined(PSIDataSchema),
   sentMessages: nullToUndefined(z.array(SentMessageSchema)).default([]),
+  refreshPSI: nullToUndefined(z.boolean()),
```
stevenelleman, line 15: I moved PullPSI from the User to the Connection.

apps/frontend/src/lib/psi/refresh.ts
```diff
@@ -0,0 +1,238 @@
...
+export const updateConnectionPSISize = async (newVerifiedIntersection: Intersection, user: User, session: Session, connection: Connection, triggerPull: boolean): Promise<void> => {
+  // set the size of intersection here
+  const intersectionSize: number = JSON.stringify(newVerifiedIntersection).length;
+
+  const newUserData: UserData = user.userData;
+  if (!newUserData.connectionPSISize) {
+    newUserData.connectionPSISize = {};
+  }
+
+  newUserData.connectionPSISize[connection.user.username] = intersectionSize;
```
stevenelleman, line 224: In a followup PR I'd like to move this out of UserData and into Connection.

apps/frontend/src/lib/storage/localStorage/user/connection/message/index.ts
```diff
@@ -148,6 +155,23 @@ async function handleEdges(session: Session, user: User, messages: EdgeMessage[]
...
+  for (const message of messages) {
+    if (message.senderUsername && user.connections && user.connections[message.senderUsername]) {
+
+      if (user.userData?.settings?.automaticPSIEnabled) {
+        // Only if the setting is enabled should PSI be automatically refreshed
+        const newVerifiedIntersection = await refreshPSI(user, user.connections[message.senderUsername]);
+
+        if (newVerifiedIntersection) {
+          // If intersection is successful, set the size of intersection
+          await updateConnectionPSISize(newVerifiedIntersection, user, session, user.connections[message.senderUsername], true);
```
stevenelleman, line 168: True is passed in here to set refreshPSI to true, so that on the connection page we'll fetch the results from the PSI.

apps/frontend/src/pages/people/[username].tsx
```diff
@@ -270,6 +262,23 @@ const UserProfilePage: React.FC = () => {
             setShowCommentModal(true);
           }
 
+          // If refreshPSI set to true, run refreshPSI and get most recent verified intersection
+          if (user && user.connections && user.connections[username]) {
+            const connection = user.connections[username];
+
+            if (connection.refreshPSI) {
```
Comment (stevenelleman) [inline apps/frontend/src/pages/people/[username].tsx:269]: Here's the order of steps: 
1. Receive PSI message for connection X  
2. Refreshes PSI for connection X 
3. Set connection X's refreshPSI to true 
4. When we next navigate to X's page, pull the results by refreshing (we can't pull directly because there's no guarantee the PSI results will still exist in server memory). 

Currently (2) is a redundant refresh so I think I may remove it (but keep (3) so the second refresh still works).

apps/frontend/src/lib/psi/refresh.ts
AndrewCLu: yup makes sense

apps/frontend/src/lib/storage/types/user/userData/index.ts
```diff
@@ -5,12 +5,29 @@ import { TelegramDataSchema } from "./telegramData";
...
+  ConnectionPSISizeSchema,
+} from "@/lib/storage/types/user/userData/psiSizeData";
+
+export const UserSettingsSchema = z.object({
+  automaticPSIEnabled: nullToUndefined(z.boolean().nullable()),
+
+  // NOTE: In a future PR swap tapGraphEnabled to live in settings
+  // tapGraphEnabled: nullToUndefined(z.boolean().nullable()),
+
+  // NOTE: currently dark theme lives in localstorage at `cursive-connections-theme` key -- should it be moved to settings to persist it between sessions?
```
AndrewCLu, line 27: yes, in future

apps/frontend/src/pages/profile/index.tsx
```diff
@@ -369,6 +373,15 @@ const ProfilePage: React.FC = () => {
                 }}
               />
             </div>
+            <div className="py-0">
+              <ToggleSwitch
```
AndrewCLu, line 377: Can we have a log here to see how many people turn this on?

AndrewCLu (approved):
LGTM, minor comments

PR #251, opened 2024-12-11
Sanitize Newline Escape Sequence

stevenelleman:
I sanitized: 
- escape sequences^1
- comments in csv^2
- special characters in csv^3

When it came down to it only newline characters (`\n`) messed up exports, I wasn't able to intentionally mess up the exports by using other escape sequences (e.g. `\r`, `\t`), comment (e.g. `#`), and special characters (e.g. `,` , `"`, `\`)  

This works for all my contacts but we could identify other missing edge cases that are outside the set of my ~190 connections. 

Sources: 
1: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_escape, https://docs.oracle.com/javase/tutorial/java/data/characters.html, https://www.geeksforgeeks.org/what-are-escape-characters-in-javascript/ 
2: https://fastcsv.org/guides/examples/handle-comments/
3: https://support.aodocs.com/hc/en-us/articles/5466301293083-Special-characters-in-CSV-files-for-migrations-from-scratch#-371457883

### DIFF
--- a/apps/frontend/src/lib/exports/index.ts
+++ b/apps/frontend/src/lib/exports/index.ts
@@ -9,7 +9,13 @@ function sanitizeCSVExport(data: string | undefined): string {
     return "";
   }
   data = data.replaceAll(";", ",");
-  return data.replaceAll("\n", " ");
+  data = data.replaceAll("\n", " ");
+  data = data.replaceAll("\b", " ");
+  data = data.replaceAll("\f", " ");
+  data = data.replaceAll("\r", " ");
+  data = data.replaceAll("\t", " ");
+  data = data.replaceAll("\v", " ");
+  return data
 }
 
 function flattenAndEscapeUserData(user: Connection): FlattenedUserData {
--- a/apps/frontend/src/lib/exports/index.ts
+++ b/apps/frontend/src/lib/exports/index.ts
@@ -4,29 +4,30 @@ import { toast } from "sonner";
 
 // NOTE: the CSV delimiter is a semicolon. If a user-inputted value has a semicolon it's swapped out with a comma to
 // prevent the exported CSV from being corrupted
-function swapSemicolons(data: string | undefined): string {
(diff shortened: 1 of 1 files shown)


apps/frontend/src/lib/exports/index.ts
```diff
@@ -4,29 +4,36 @@ import { toast } from "sonner";
...
     return "";
   }
-  return data.replaceAll(";", ",");
+  data = data.replaceAll(";", ",");
+  data = data.replaceAll("\n", " ");
+  data = data.replaceAll("\b", " ");
+  data = data.replaceAll("\f", " ");
+  data = data.replaceAll("\r", " ");
+  data = data.replaceAll("\t", " ");
+  data = data.replaceAll("\v", " ");
```
stevenelleman, line 17: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_escape

apps/frontend/src/lib/exports/index.ts
```diff
@@ -4,29 +4,43 @@ import { toast } from "sonner";
...
+  // Delimiter
+  data = data.replaceAll(";", ",");
+
+  // Escape Characters
+  data = data.replaceAll("\n", " ");
+  data = data.replaceAll("\b", " ");
+  data = data.replaceAll("\f", " ");
+  data = data.replaceAll("\r", " ");
+  data = data.replaceAll("\t", " ");
+  data = data.replaceAll("\v", " ");
```
stevenelleman, line 20: I tried to use a single regex for all these options but it messed up the exports. Will keep looking.

stevenelleman: Nit: I should change the comment to *Escape Sequences

AndrewCLu (approved):

